Issue #8, opened 2020-03-01
Filter with dropdowns


jshom:
- be able to select sub semester courses
- be able to select courses by department

PR #9, opened 2020-03-01
Filter by dept & sub-semester

CoreyBuckley:
This PR implements the feature described in #8. I have also added a default 'All' option for both department and subsection dropdowns.

I had to add the date_start and date_end fields in the SQL for what was being returned from `/api/class`.

Because of this, those fields had to be added to the GROUP BY clause which means "duplicate" courses are returned (at least from a user perspective). These duplicate courses are those that have multiple start and end dates. As of the `/rpi-data/summer-2020.csv` file, these are those courses:

```
 department | level
------------+-------              
 ENGR       |  2250  
 STSS       |  4100 
 ENGR       |  4010 
 ECON       |  4230 
(4 rows)
```

Each one of these has a 'duplicate' that will appear in the UI. Previously, `get_classes_full` returned 134 records. Now 138 are returned.

I'm not sure how this should be handled, so, I'll defer. But, filtering does work.

### DIFF
--- a/src/db/classinfo.py
+++ b/src/db/classinfo.py
@@ -25,6 +25,8 @@ class ClassInfo:
               c.level,
               concat(c.department, '-', c.level) as name,
               max(c.title) as title,
+              c.date_start,
+              c.date_end,
               json_agg(
                 row_to_json(section.*)
               ) sections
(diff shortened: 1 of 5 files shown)


jshom:
Ye we'll take a look at the duplicates within semesters next week. For now this should work :+1:. Also, we'll need larger backend filter on semester, this is not important right now but this will come up in the future. i.e. filter on classes full and class info per semester.

Issue #17, opened 2020-03-03
Distribution — RPI:Parsing Errors


jshom:
**tp column fix**
We are ignoring courses that include the TP column
example course:
- ADMIN-1824

**change parsing method**
do not rely on length, ignoring courses that are not of length 5 and 11, take into account all fields and edge cases.

marchdan:
Working to make parser more robust. Will post example of new parse when completed

marchdan:
Snippet from new CSV
`
STUDENT SUCCESS LABS | LEC | 0 | W | 1:00PM | 1:50PM | Trzepacz | 120 | 0 | 120 | ADMN | 2020-05-26 | 2020-08-21 | SUMMER 2020 | 15486 | 1824 | 1
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
SUMMER ADMINIST REGISTRATION | LEC | 0 |   | **TBA** |   | Grega | 625 | 7 | 618 | ADMN | 2020-05-26 | 2020-08-21 | SUMMER 2020 | 14908 | 6600 | 1
MATERIALS AND DESIGN | LEC | 2 | R | 10:30AM | 12:35PM | Staff | 12 | 2 | 10 | ARCH | 2020-05-26 | 2020-08-21 | SUMMER 2020 | 15343 | 2510 | 1
ARCHITECTURAL DESIGN STUDIO 5 | LEC | 5 | MWR | 2:00PM | 5:50PM | Staff | 32 | 32 | 0 | ARCH | 2020-05-26 | 2020-08-21 | SUMMER 2020 | 15413 | 4770 | 1
MASTER'S THESIS | LEC | 1-12 |   | **TBA** |   | Staff | 10 | 0 | 10 | ARCH | 2020-05-26 | 2020-08-21 | SUMMER 2020 | 15031 | 6990 | 3
MASTER'S THESIS | LEC | 1-12 |   | **TBA** |   | Staff | 0 | 0 | 0 | ARCH | 2020-05-26 | 2020-08-21 | SUMMER 2020 | 15057 | 6990 | 4
ARCHITECTURE DISSERTATION | LEC | 1-15 |   | **TBA** |   | Figueiro | 10 | 0 | 10 | ARCH | 2020-05-26 | 2020-08-21 | SUMMER 2020 | 15240 | 9990 | 1
[...]

PR #20, opened 2020-03-04
CSV parse fix

marchdan:
In reference ot #17 This will parse out all courses into the CSV. I am opening a pull requst so @CoreyBuckley Can look at the new parsed CSV for any minor changes which need to be taken into account, such as courses with no type or  no start/end time. Please suggest any additional changes which need to be made


marchdan:
Fixed issues should now line up with #17

PR #37, opened 2020-03-06
Add course_location to course_session schema & CSV parsing

### LINKED ISSUE #36: Distribution — RPI:Update course schema to include location
In reference to #28, we need the database to support the location of the class when it becomes available.

CoreyBuckley:
- Adds location field to table `course_session`.
- Very basic validation to make sure csv file uploaded when you submit
- Back btn on admin dashboard page
- Use location in the ICS schedule events

Closes #36


123joshuawu:
Location should probably be tied to course session, considering that lab and lectures can be in different locatinos

jshom:
Yep, I agree with @123joshuawu, this should be in the course session table

CoreyBuckley:
@123joshuawu @jshom I agree, thanks for the feedback.

PR #50, opened 2020-03-13
Add MIT license to external ICS lib

CoreyBuckley:
While it's possible to implement the ICS standard, I have found the documentation kind of hard to follow at times. So, unless the group feels they don't want to use an external lib for this, I'm giving proper attribution as per the MIT license of this code.


jshom:
Yep thanks for including this!

PR #51, opened 2020-03-13
Add SubsemesterScheduler, component reorganization, Prettier code style, add jsdocs

123joshuawu:
TODO: 
- [X] review ICS calendar changes, idk if it works @CoreyBuckley 
- [x] more thoroughly test subsemester functionality
- [ ] documentation generation for sfc using jsdoc annotations

Two main aspects of the changes below are
1) Added `SubSemesterScheduler` for toggling schedules based on subsemester
2) Splitting `CourseList` and `SelectedCourses` into separate components

Additional noteworthy changes:
* Added proxy to Vue devServer so CORS no longer required in backend
* Added prettier for enforcing code style
* Added jsconfig.json for vscode intellisense import of '@/' webpack alias
* Added jsdoc.config.json for jsdoc documentation generation

Overview of `SubSemesterScheduler`
`SubSemeseterScheduler` holds a list of `Schedules` each associated with a `SubSemester`. When adding `courseSections` the `SubSemesterScheduler` determines which `SubSemester` the `course` affects using `withinCourseDuration` and then adds the `courseSection` to the affected `Schedule`. For removal, atm every `Schedule` attempts to remove the `courseSection`. Since the number of comparisons are relatively low and each operation is cheap, the approach should work fine. Additionally, `SubSemesterScheduler` also adds the affected `subsemester` to the `ScheduleConflict` error.

Part of this PR was to ensure the `Schedule` had no concept of `subsemesters`, keep it such that a `Schedule` only operates with `courseSections`. My thinking is for `Schedule` to only be concerned with `ScheduleConflict` between `courseSections`. Other concepts like `courses` or `subsemesters` or `semesters` should be handled outside of `Schedule`.

Overview of Front-end Components:
The basic design is `Main` contains various components that each have their own function. Any state that involves more than one child component is hoisted to `Main` and children components make changes through `events`. This can be seen in `selectedCourses` which is affected by/affects `CourseList`, `SelectedCourses`, `Schedule`s. The management of `Schedule`s could be moved to a separate module along with the `scheduler`, but for now that functionality resides in `Main`. 

Additionally, an attempt was made to move common functionality into reusable modules. `ColorService` was made into a singleton. Mixins were created for `TimeDateMixin` and `Notifications`, however after further research, it is probably better to move away from mixins. 
Honestly im rambling im tired lol


src/yacs-web/src/components/ScheduleEvent.vue
```diff
@@ -22,19 +22,19 @@ export default {
 }
 </script>
 
-<style scoped lang="scss">
```
CoreyBuckley: Is there a reason for removing the `scoped`?

src/yacs-web/src/components/SelectedCourses.vue
```diff
@@ -0,0 +1,153 @@
...
+                this.$emit('removeAllCourseSections', course);
+                this.$emit('addCourseSection', course, section);
+            }
+        },
+        exportScheduleToIcs () {
+            let calendarBuilder = window.ics()
+            let semester;
+
+            for (const course of Object.values(this.courses)) {
+                for (const section of course.sections.filter(s => s.selected)) {
```
CoreyBuckley: As of now, this will give back only 1 course since you can only select 1 section as per the `toggleCourseSection`.

I like the idea of "future-proofing" code. But not sure if this would cause some confusion since it might lead one to believe you can select multiple sections of the same course.

(_Note:_ I believe in old yacs TAs wanted the ability to add multiple sections of the same course. Not sure what the verdict on that was.. )

src/yacs-web/src/components/SelectedCourses.vue
```diff
@@ -0,0 +1,153 @@
...
+                // Only allow selection of one section per course
+                this.$emit('removeAllCourseSections', course);
+                this.$emit('addCourseSection', course, section);
+            }
+        },
+        exportScheduleToIcs () {
+            let calendarBuilder = window.ics()
+            let semester;
+
+            for (const course of Object.values(this.courses)) {
```
CoreyBuckley: The reason I turned `this.courses` into a dictionary was so that it was easy and fast to get the course information of a given session.

However, when looping over all courses and figuring out the matching section, it kind of eliminates the need for `courses` to be a dictionary. `_getCourseIdentifier` doesn't really serve a purpose anymore. I struggle to think if there would be an eventual need.

src/yacs-web/src/components/SelectedCourses.vue
```diff
@@ -0,0 +1,153 @@
...
+            //         let courseInfo = this.courses[session._courseKey];
+            //         semester = session.semester;
+            //         calendarBuilder.addEvent(`Class: ${courseInfo.title}`, "LEC day", session.location, new Date(`${courseInfo.date_start.toDateString()} ${session.time_start}`), new Date(`${courseInfo.date_start.toDateString()} ${session.time_end}`), {
+            //             freq: "WEEKLY",
+            //             interval: 1,
+            //             until: courseInfo.date_end,
+            //             byday: [this.ICS_DAY_SHORTNAMES[session.day_of_week]]
+            //         });
+            //     }
+            // }
```
CoreyBuckley: Can remove old implementation.

Originally I had thought each individual schedule would produce an ICS file pertaining to the events of only that subsemester. But grouping all the subsemesters into one ICS file makes sense.

src/yacs-web/src/pages/Main.vue
```diff
@@ -5,110 +5,85 @@
...
+                    @addCourse="addCourse"
+                />
 
             </b-col>
             <b-col md='8'>
-                <Schedule :courses="selectedCourses" :courseIdentifierFunc="_getCourseIdentifier" @unselectCourse="removeCourse"></Schedule>
+                <b-form-select v-model="selectedScheduleIndex" :options="selectedScheduleOptions"></b-form-select>
+   
+                <Schedule
+                    v-for="(schedule, index) in scheduler.schedules"
```
CoreyBuckley: Is it necessary to make separate Schedule components? Makes sense as a mental model, but you can probably get away with one instance and do:

```
                <Schedule
                    v-if="scheduler.schedules[selectedScheduleIndex]"
                    :schedule="scheduler.schedules[selectedScheduleIndex]"
                />
```

src/yacs-web/src/services/ScheduleService.js
```diff
@@ -8,23 +8,33 @@
...
+ * @property {string} id
  */
 
+ /**
+  * @typedef {Object} Subsemester
+  * @property {Date} date_start
+  * @property {Date} date_end
+  * @property {string} date_start_display
+  * @property {string} date_end_display
+  * @property {string} text
```
CoreyBuckley: More descriptive field name would be nice, e.g. `displayString`

CoreyBuckley:
Honestly, very impressive. Nice job.

I have one maybe controversial thought:

If I have one 12-week course and one 6-week course, then there's bound to be overlap. However, if I look at that 12-week subsemester, I will not see the 6-week course.

E.g. 

**12-week view**

image

**6-week view**

image

COMM-2520 doesn't show up in the 12-week view. Should it? IMO, feel like it makes logical sense for it to appear, but I can see why you don't show it.

I'm starting to feel like a date range / configurable timeline makes most sense from a user perspective.

image

Future feature? 😅

src/yacs-web/src/pages/Main.vue
CoreyBuckley: I have pushed this change to your branch because I could not find a problem with this. Feel free to revert if you know of an issue.

src/yacs-web/src/components/SelectedCourses.vue
123joshuawu: Good point, this will probably also change after the addition of 'possible schedules' based on selected course sections. But in the meantime I agree and will remove that constraint.

src/yacs-web/src/components/SelectedCourses.vue
123joshuawu: I like the concept of generating ids and the dictionary and proposed having this done on the backend. At the moment, since the data is nested, access to the children elements are generally within the scope of the parent, so the backpointer's are usually not necessary. However, the id is definitely useful for normalized data (no nesting) which is still an option we may explore in the future. 
Additionally, object properties now are ordered in modern browsers + ES6 so using them as arrays should be fine. So we can probably leave this as it is and in the case we need to use them or decide to normalize the data they will come in handy.

src/yacs-web/src/pages/Main.vue
123joshuawu: Probably insignificant in this use case but my understanding is that `v-if` is conditional rendering while `v-show` is just visibility toggling. So `v-if` will create/destroy the component on every change, but initially will only load one component so "light load, heavy toggle". On the other hand, `v-show` will create and all the components to the DOM on initial load and changes the `display` of components on change so "heavy load, light toggle".
In this case, I'm making the assumption that users are probably going to toggle quite often, so I opt for `v-show`. In my mind, I would probably use `v-if` for say rendering different pages for user and admin. 
The other aspect of it is trying to keep a one-to-one pairing between `ScheduleComponent` and `ScheduleService`. I think this will help with data binding and reactivity but those are my thoughts and I can't come up with a good explanation on why rn lol.
Link to docs: https://vuejs.org/v2/guide/conditional.html#v-if-vs-v-show

src/yacs-web/src/services/ScheduleService.js
123joshuawu: Agree will do

src/yacs-web/src/pages/Main.vue
CoreyBuckley: @123joshuawu Yes, you are correct. I originally threw that on there because the schedule service object was undefined and throwing errors. It would be better to remove that v-if and probably just put a ternary inside the :schedule. Feel free to make that change 🙂

123joshuawu:
Yeah I like the idea of being able to see stuff on one schedule. 
One of the designs during the brainstorms that I really liked was something like this
Screen Shot 2020-03-14 at 2 59 18 PM
Hopefully just by looking at it it should be pretty clear which subsemester each course belongs to.
The left and right halves are two variations on the spacing/border placements. But the main idea was to show all the courses in one schedule and have half blocks for each subsemester.
But we eventually went with the toggle.

src/yacs-web/src/pages/Main.vue
123joshuawu: Hm interesting did you find out what was causing the service to be undefined?
This is also a sign of my lack of error catching throughout the project atm lol

CoreyBuckley: image

I believe it happens because, in Main.vue, the SubSemesterScheduler is unpopulated with the schedule services by the time `created()` is called on the Schedule component (which depends on the schedule service getter `courseSessionsOnDay`.

While the `created` lifecycle method is synchronous, you're calling `getSubsemester` which is a promise and so that won't halt the Vue lifecycle. Thus, when Schedule.vue's created() is called, its schedule service prop is undefined since it depends on that getSubsemester promise in the parent.

Relevant links: 

- [Async code in created](https://forum.vuejs.org/t/cant-get-created-hook-to-work-when-its-async/28604/4)
- [Parent and child component lifecycle](https://medium.com/@brockreece/vue-parent-and-child-lifecycle-hooks-5d6236bd561f)

src/yacs-web/src/components/ScheduleEvent.vue
123joshuawu: I'm guessing there's something weird going on between the inline styles and the scoped styles but removing it makes it look correct

CoreyBuckley:
I like the idea, I just think there's still a possibility to misinterpret it. A user misinterpreting the distinction of subsemester possibly means they attend a non-going class. 😅

I think if you included the start date on the schedule event for those courses in that later 6-week half it'd be much more obvious.

Example:

image

Other than that, I think I prefer it TBH.

src/yacs-web/src/pages/Main.vue
123joshuawu: Interesting... I guess in my head I imagined that the Schedule component would only be created by `v-for` when `SubSemesterScheduler` adds a new `ScheduleService` but I guess I'm missing something

CoreyBuckley: My bad, I should have been more specific. I was referring to when `v-for` wasn't on the Schedule component. You are correct when there is a `v-for` attached to it.

As for having multiple schedule components vs one, I didn't think the binding would be that expensive. Feel like the most expensive part is the re-rendering, but the ScheduleEvent is absolutely positioned so it shouldn't be that expensive.

Regardless, it hopefully shouldn't matter much either way. 🤷‍♂

123joshuawu:
Yeah the only thing is if the day came and RPI decided to do three subsemesters instead of two, then we'd have to redesign again lol.

src/yacs-web/src/pages/Main.vue
123joshuawu: Yeah, marking as resolved for now, can open other threads for continuing this line of thought later

jshom:
This, is, so, cool!
Just one thing, seems when I click a course the event is not filling in with any info. Can you add class.title and class.section. thx!
Screen Shot 2020-03-15 at 1 48 38 AM

jshom:
Also, the whole semester (05/25-08/20) should not be an option in the dropdown. There are in practice two parts of the semester. We can defer this to a solution where in the admin panel we select which date ranges map to which "semester parts" then we can show semester part.

So example:
date range (05/25-08/20) belongs to part1 and part 2
date range (07/12-08/20) belong only to part 2

What you think?

123joshuawu:
@jshom Yeah I like the use of 'semester parts' over date ranges, that's a really good idea.

As for the event not showing info, I had the issue a little back but it should be fixed by now, make sure you have all the recent commits.
Displaying the course information is going to be a little problematic right now just because the `Schedule` and its child `ScheduleEvent` only have access to the `CourseSession`. I'm planning on moving `Course`, `CourseSection`, and `CourseSession` to centralized store some time soon which will fix this issue, but for now, it just shows crn and section number.

Issue #21, opened 2020-03-05
Distribution — Select Semester


jshom:
Job:
When a new semester comes we want to be able to change to a new semester without restarting the application.

Goals:
In admin panel, add a section for default semester. Drop down with options from a vial able semesters in db

jshom:
Note:
This should change the schedules viewable from the main page

PR #56, opened 2020-03-14
Scheduling switching

marchdan:
A preliminary draft of semester switching. Cannot be completed until multi-semester capability is added to the db.


jshom:
Regarding #21.
Lets
1. Create settings section in admin panel for the semester select. `POST /setting`
2. `GET /setting` route to share the selected default semester from the admin panel. Then set the default on frontend using the setting default
3. Fix formatting here 
Screen Shot 2020-03-14 at 7 52 19 PM

Looks good man!

Issue #34, opened 2020-03-06
Scheduling — Class List, scroll class list independent of week schedule


jshom:
Change the css for scrolling in the class list so that the schedule render is always in view

PR #61, opened 2020-03-15
Ui tweaks sortorder space optimization

jshom:
Screen Shot 2020-03-15 at 1 16 21 AM
Contribute to #34, need better veritical page partitioning when we are more done with functionality as it is now set to 700 px


123joshuawu:
Maybe good to remove the sorting of results from YacsService and leave that to the backend? Can do the same for Departments, Subsemesters, and possibly CourseSections and CourseSessions?

jshom:
Agree yep let me double check

jshom:
ah yep, forgot to commit that after merge. shifting sorting to db in a sec, 2 min.

CoreyBuckley:
I will be committing some minor changes:

- Don't need overflow-x to always be scroll.
- Make `height` max-height instead so when I use mobile view I don't have a large chunk of empty space if I were to filter to one course.

PR #62, opened 2020-03-15
Cross plat start script

CoreyBuckley:
Start both frontend and backend via `npm run start`. Should be able to run this command anywhere in the project I believe.

Uses `npm-run-all` which allows for running script entries in the `package.json` concurrently and combines the output of both in the terminal.

(I recommend installing npm-run-all globally, e.g. `npm i -g npm-run-all` )


jshom:
`npm i && FLASK_APP=src/app.py npm start` to start

PR #64, opened 2020-03-17
Semester switching enabled

marchdan:


CoreyBuckley:
Some margin below the header would be good, ``<Header class='mb-3' />``.

And exposing the selected semester to to the parent component, Main.vue would be good.

marchdan:
@CoreyBuckley I will add the space below the header.

The selected semester is in the return data, where in header should I expose it? Header is being imported into Main already

123joshuawu:
I was thinking for this what if we used router-view for different semesters.

marchdan:
@123joshuawu I am impartial to what we use, as long as I am able to access a list of semesters, and I know where I need to indicate the current semester

jshom:
Sweet!
Two things:
1. If I change "SUMMER 2020" to "FALL 2020", my view does not appear to change
2. We need to be able to set default from admin panel as per #21

src/yacs-web/src/components/Header.vue
```diff
@@ -1,7 +1,7 @@
 <template>
     <b-navbar type="light" variant="light">
         <b-navbar-brand class = "logo" href="#">YACS</b-navbar-brand>
-
+        
 		<b-form-select label= "semester" class="semesters" v-model="selectedSemester" :options="semesterOptions">
```
123joshuawu: if you add a `text-field="semester"` and `value-field="semester"` to the `b-form-select.semesters`, then in `created ()` you can just push `data` returned by `getSemesters` instead of creating a new array.

src/yacs-web/src/pages/Main.vue
```diff
@@ -1,8 +1,8 @@
 <template>
   <b-container fluid class="pt-3">
+    <Header class='mb-3'></Header>
```
123joshuawu: I think you need to register a listener for the `@new-semester` event here

marchdan:
@jshom The view doesn't change because we don't have that ability yet. The different semesters aren't distinct sets of data.
Where in the admin panel is the default stored? I will just access it.

src/yacs-web/src/components/Header.vue
marchdan: I am just mimicking all of the other similar methods.

src/yacs-web/src/pages/Main.vue
marchdan: How would I add a listener there? What am I listening for?

jshom:
You'll need to add a new section in the /admin panel, this is a part of the issue to create that section and set the default.

marchdan:
@jshom So I should add a POST method to AdminService for the default semester?

jshom:
Yes and on the python backend as well

src/yacs-web/src/pages/Main.vue
123joshuawu: So in `Header.vue` you're `watch`ing `selectedsemester` and when the value changes you call `this.$emit('new-semester', val)`. So, when the value of `selectedsemester` changes, `<Header>` will emit the event `'new-semester'` with the new value. So to update `currentSemester` in `Main.vue`, just add `@new-semester="newSemester"` to your `<Header>` tag. An example of this is `<CourseList @addCourse="addCourse" />

src/db/admin.py
```diff
@@ -0,0 +1,21 @@
...
+
+	def get_semester_default(self):
+		return self.db_conn.execute("""
+			select
+				semester
+			from
+				admin_setting
+		""", None, True)
+
+	def set_semester_default(self, sem):
```
jshom: `sem` -> `semester`, we have plenty horizontal space, for the rest of the codebase lets aim to be as explicit as possible in naming, easier to reason about.

src/yacs-web/src/pages/Admin.vue
```diff
@@ -15,21 +15,32 @@
         <input type="Submit" class="btn btn-success btn-sm" value="Submit" />
       </form>
     </section>
+    <section id="select-data">
```
jshom: select-semester, data is ambiguous

marchdan: The idea is that there will be multiple data items to select, not just the default semester.

src/app.py
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/python3
```
123joshuawu: This file has been moved to `yacs.n/src/api/app.py`

marchdan:
Communications and setting should work. Need to add semester sorting to course display

scripts/dev.sh
```diff
@@ -12,7 +12,7 @@ echo "Starting @ $(date)" > $PATH_ROOT/logs/dev.txt
 
 # run flask
 cd src/api/
-flask run --reload >> $PATH_ROOT/logs/api.txt 2>&1 & # run in bg and port logs
+DB_USER=marchdan DB_PASS=" " flask run --reload >> $PATH_ROOT/logs/api.txt 2>&1 & # run in bg and port logs
```
123joshuawu, line 15: Let's not add this to master lol. we need .env files

src/app.py
123joshuawu: bump

src/web/src/components/CourseList.vue
```diff
@@ -84,13 +101,15 @@ export default {
...
-      return this.courses.filter(({ date_start, date_end, department, title }) => {
+      return this.courses.filter(({ date_start, date_end, department, title, semester }) => {
         return (
           (!this.selectedSubsemester ||
             (this.selectedSubsemester.date_start.getTime() === date_start.getTime() &&
               this.selectedSubsemester.date_end.getTime() === date_end.getTime())) &&
           (!this.selectedDepartment || this.selectedDepartment === department) &&
-          (!this.textSearch || title.includes(this.textSearch.toUpperCase()))
+          (!this.textSearch || title.includes(this.textSearch.toUpperCase())) &&
+          (!this.selectedSemester || this.selectedSemester ===
```
123joshuawu, line 111: Seems that the default value for `selectedSemester` is not a falsey value so this actually breaks when there is no defaultSemester

src/web/src/components/Footer.vue
```diff
@@ -7,10 +7,14 @@
...
             <!-- TODO: Autogenerate these when doing the user side, semester select -->
             <strong class="section-head">Other Semesters</strong>
-            <a class="link" href=#>Fall 2019</a>
-            <a class="link" href=#>Fall 2020</a>
-            <a class="link" href=#>Spring 2020</a>
-            <a class="link" href=#>Summer 2020</a>
+            <a 
+              v-for="semester in semesterOptions"
+              v-bind:key="semester.text"
+              v-bind:value="semester.value"
```
123joshuawu, line 13: 1) `:` is shorthand for `v-bind:`
2) does `<a>` need a `:value` prop?

src/web/src/components/Footer.vue
```diff
@@ -40,15 +44,24 @@
...
 <script>
 
+import { getSemesters } from '@/services/YacsService';
+
 export default {
     name: 'Footer',
     data() {
       return {
-        field: 0
+        field: 0,
```
123joshuawu, line 53: is this value being used?

scripts/dev.sh
marchdan: XD I will fix that

src/web/src/components/CourseList.vue
marchdan: Fixed

src/web/src/components/Footer.vue
marchdan: There is an error produced if :value is not given :(

src/web/src/components/Footer.vue
marchdan: No, but I didn't add it...

src/api/app.py
```diff
@@ -60,6 +60,20 @@ def get_semesters():
...
+
+@app.route('/api/defaultsemesterset', methods=['POST'])
+def set_defaultSemester():
+    info = request.get_json()
+    success, error = admin_info.set_semester_default(info['default'])
+    if success:
+        return Response(status=200)
+    else:
+        print(error)
+        return Response(error.__str__(), status=500)
```
smgibson19, line 76: Is this new code setting the default semester or allowing it to be changed? I want to understand what is happening better.

marchdan: The bottom route is calling set_semester_default with the given default semester in order to set the default semester. 

The top rout simply retrieves the current default semester

PR #65, opened 2020-03-18
Map date range to name.

### LINKED ISSUE #63: Scheduling — Semester Parts
Currently we have date ranges in the dropdown for partial semesters, this does not make perfect sense as in a semester with 3 date ranges (full, first half, second half) during a semester, there would only really be two separate schedules. 

To solve this, lets create a mapping from date range -> semester part.
Example:
Course (05/25-08/20) is in semester parts 1 & 2
Course (05/25-07/10) is in semester part 1

We will need to be able to manipulate the scheduling mappings in the admin panel.
We should able to:
1. Add a date range mapping (name and date range)
2. Edit a date range mapping (name and date range)
3. Ignore a date range mapping (ignore flag)

This should reside in a table date_range_semester_part that will have a wrapper in the src/db/ wrappers directory.
NOTE: add the `create table` statement in the src/data/schema/ directory and add the script to execute in the src/data/build.sh

At completion, in the example of RPI's summer semester, we should be able to see summer part 1, summer part 2 in the dropdown if someone selects a partial semester class. If all selected classes are full semester, then only show the full summer semester.

CoreyBuckley:
- [X] Add `semester_date_range` tbl to schema
(No auto populate on tbl, currently manual)
- [X] Add wrapper to `db` for the tbl
- [X] Create
- [X] Read
- [X] Update

Main functionality is there. Can update look/feel later.

- Will require rebuilding schema.
- Split admin pages into sep routes for now. Access at /admin/mapdate
- Can update multiple date bindings at a time, all in one form that is submitted.
- Visual indication of success / failure on button via color change.
- Fix dates being off due to GMT/UTC in DB vs JS auto converting to local timezone of EST for subsemesters, courses which affected CourseList, date mapping, and the export to ICS.
- Check for error on flask endpoints before returning
- There are no default name bindings for the date ranges. getSubsemesters() from YacsService instead checks if there is a valid DB value and uses that, else use the [date start] - [date end] name

Closes #63


src/app.py
```diff
@@ -78,7 +61,24 @@ def uploadHandler():
         print(error)
         return Response(error.__str__(), status=500)
 
-
+@app.route('/api/mapDateRangeToSemesterPart', methods=['POST'])
+def map_date_range_to_semester_part_handler():
```
123joshuawu: Maybe sending semesterparts as a list of objects could work as well? Seems like it works

src/db/semester_date_mapping.py
```diff
@@ -0,0 +1,39 @@
...
+                        ;
+                    """, {
+                        "SemesterPartName": names[i],
+                        "DateStart": start_dates[i],
+                        "DateEnd": end_dates[i]
+                    }, isSELECT=False)
+                    if (error):
+                        return (False, error)
+                    else:
+                        return (True, None)
```
123joshuawu: I think this exits early in the case of updating multiple semester parts

src/yacs-web/src/pages/MapDates.vue
```diff
@@ -0,0 +1,142 @@
...
+        standardDate,
+        resetIndicators() {
+            this.fail = false;
+            this.success = false;
+        }
+    },
+    created() {
+        getSubSemesters().then(subsemesters => {
+            this.subsemesters = subsemesters;
+            this.standardSemesterNames = new Set(subsemesters.map(subsemester => subsemester.semester_name));
```
123joshuawu: Particular reason for using a `Set` for `standardSemesterNames`?

src/yacs-web/src/utils.js
```diff
@@ -28,6 +28,34 @@ export const readableDate = date => {
   return `${date.getMonth() + 1}/${date.getDate()}`;
 };
 
+/**
+ * Formats a `Date` into postgres date format
+ * @param {Date} date
+ * @returns {string} string in the format `YYYY-MM-DD`
+ */
+export const standardDate = date => {
```
123joshuawu: I'm stil working on getting to refactoring some stuff to use `momentjs` but can also use `momentjs` here, since its already installed 😄 , just a reminder, no need to change

src/yacs-web/src/services/YacsService.js
```diff
@@ -70,14 +70,13 @@ export const getDepartments = () =>
...
+      subsemester.date_start = localToUTCDate(new Date(subsemester.date_start));
+      subsemester.date_end = localToUTCDate(new Date(subsemester.date_end));
       subsemester.date_start_display = readableDate(subsemester.date_start);
       subsemester.date_end_display = readableDate(subsemester.date_end);
-
-      subsemester.display_string = `
-                    ${subsemester.date_start_display} - ${subsemester.date_end_display}
-                `;
+      // Used to determine what semester the subsemester is part of
+      subsemester.semester_name = subsemester.parent_semester_name;
```
123joshuawu: If can update `src/typedef.js` would be mch appreciated

CoreyBuckley:
> I like this very much lol, really great work

Thank you 😄  

> 1. Do we have an expected behaviour if the user tries to give the same name to multiple semesters? e.g. gives a warning on submit? or acceptable and works?

The user will be allowed to give the same name to multiple semesters because the PK on the semester_date_range table is (date_start, date_end). I figured 'Summer Part 1' or something should be able to be reused across multiple semester instances. I also now realize I should probably include the year in the date range UI.

> 2. The time zone thing is actually an interesting problem, really good catch btw. I'm thinking one option could be let the admin input the timezone along with the courses. Should we even consider the possibility of course times being in different time zones? In terms of the storage, I'm thinking store all datetimes as UTC+0 and then convert it on the front end when displaying to the local time zone?

I'm not sure what the full technical details are, but even though course.date_start/end in the DB is a `date`, it does come back with that midnight time attached at UTC+0 (get_subsemester or any other relevant DB operations with the python driver). 
image

Might be something to do with psycopg2 or some postgres thing. 

Then YacsService uses that date_start/date_end in new Date() which means it's already being auto converted to the local timezone.

If you're talking about letting an admin specify the timezone for course session start/end times, which is currently timezone agnostic (it's just a string, no intrinsic timezone), no idea. When a user accesses the site from a non-EST system, should the times be converted? I wouldn't think so, probably best to just put "2:00 PM EST" instead. So.. Changing the `course_session.time_start` to DB type `timestamptz` (which will store it in UTC with the EST/native offset). Will the should the DB always use the native timezone? Well, if we had distributed DB servers, then they should manually be specified. So, then that needs to be supplied in the CSV or be some manual setting.

tl;dr it might be a good idea if we want the application to scale.

src/yacs-web/src/pages/MapDates.vue
CoreyBuckley: I added the property `subsemester.semester_name` which will either be `SUMMER|FALL|SPRING {YEAR}`. Date range to name inputs are grouped under these possible semester names. So, I make sure there's no duplicate entries for these possible semester names, otherwise in the UI you'd see ("SUMMER 2020" [all date ranges pertaining to this semester]) x (repeat count)

src/db/semester_date_mapping.py
CoreyBuckley: you're right. pushed.

src/yacs-web/src/pages/MapDates.vue
123joshuawu: OH its semester_name I'm dumb lol. Good solution

123joshuawu:
What if we added a settings thing through which the user can configure what timezone they want all the times to be in? It can like default to the local time zone or something but users have the option to change manually change it to a different time zone. 
And in terms of timezone in the admin service, maybe we should save the timezone which the course information was inputted in or provide some option to set this. This is probably more of a Yacs Distribution issue, @jshom what do you think

jshom:
Looks good man!
Few things:
...
**1 - Placement**
Lets consolidate all admin components to the /admin page, need to be able to see everything there or at the least have a link. Optimally, lets show the latest semester on the admin panel and have an option to open up a modal to show this page, given it's a component shouldn't be too much work. We can toss this into another issue if you want, lmk if it is too much work to encapsulate in a modal in /admin.
**2 - Visuals**
List of fixes here:
- [ ] Placeholder in text field should not be a different color. The shade of gray is strange and upon update don't want to feel like it is not the value. Keep all standard black.
- [ ] Want to be able to update semesters individually, psychological weight of updating all semesters feels to large, want to be able to update one at a time. 
- [ ] Update Buttons - Keep the update button as disabled until the admin changes the text for a semester (then we change color and click-ability), this way I know when I changed something. Given this is an administrator using this, we want them to feel confident using the product, feedback elicits confidence. (Also rename "submit" to "update")
- [ ] Remove Paging - At most there will be 3/4/5 semester parts at most, no need to page through 5 records. Lmk if I'm missing something with regards to the decision to do paging here.

Screen Shot 2020-03-27 at 3 00 17 PM

Great work man, just final polishing stuff!

src/yacs-web/src/services/YacsService.js
CoreyBuckley: done

src/app.py
CoreyBuckley: I believe it works too. Going to resolve for now.

src/yacs-web/src/utils.js
CoreyBuckley: :+1:

Yeah, JS dates are crazy huh

CoreyBuckley:
@jshom Thanks for the feedback. I actually thought of making each semester grouping having its own submit button but wasn't sure if it was a desired thing. I definitely agree with your reasoning, good observation.

All of your requested changes have been made and pushed.

PR #66, opened 2020-03-20
Add toggle of side panel between SelectedCourses and CourseList

123joshuawu:
Move `SelectedCourses` from cards into list display and allow users to toggle between viewing `CourseList` and list of `SelectedCourses`.

Created `CourseListing` component that encapsulates list display of a course and actions that are enacted on a course. The component is designed to be put in a `list-group-item`. It displays course information on the left and some `actions` on the right. The actions shown are customizable and include add (plus), remove (minus), and toggle course section collapse display (down chevron) and the course actions emit events `addCourse` and `removeCourse` respectively. 

Added flex scroll to CourseList and SelectedCourses. Both lists now greedily consume available height in parent container AND scroll when child content overflow's height.

`CourseListing` also has an option to "lazy load" the collapsible course sections. I think I had performance issues when I first added `CourseListing` to `CourseList` and was rendering ~242 courses and all their sections and sessions. This option seemed to fix that. 

Finally, changed `ScheduleEvent` to a functional component. Since `ScheduleEvent` held no state and was essentially a wrapper for the rendering function, this fits the functional component MO. I'll see if we can use this for some other components as well but for now `ScheduleEvent` jumped out as the most apparent candidate. 

Other changes: 
* Moved ICS export to `Main.vue`
* Fixed bug in `CourseList.vue` computed `filteredCourses`. Was trying to access deprecated attribute. Probably speaks to our need for unit tests, am evaluating.  

Note:
Since `CourseListing` is the grandchild of `Main.vue` (where `courses` state resides), events are passed up to `Main.vue` using `v-on:$listeners`. Seems this is a bit of a hack; the better option is to use centralized state management. Will leave that for separate PR.


jshom:
This is awesome... good stuff man!
Quick few things:
*1/* Let's not show the dropdown for the sections when in "class search", once we have descriptions it will feel right as the job of the search is to find classes you want to take, not find a correct section.
*2/* When I select the "+" lets change that icon to "x" on-class click so we can remove the selection from the "class search" view in case I misclicked.

123joshuawu:
> This is awesome... good stuff man!
> Quick few things:
> _1/_ Let's not show the dropdown for the sections when in "class search", once we have descriptions it will feel right as the job of the search is to find classes you want to take, not find a correct section.
> _2/_ When I select the "+" lets change that icon to "x" on-class click so we can remove the selection from the "class search" view in case I misclicked.

PAtched

PR #67, opened 2020-03-22
Add crn check between courses and sections to fix duplicate sections

123joshuawu:
Not sure if there are other side effects but seems to fix the duplicate sections

### DIFF
--- a/src/db/classinfo.py
+++ b/src/db/classinfo.py
@@ -53,7 +53,8 @@ class ClassInfo:
             ) section
             on
               c.department = section.department and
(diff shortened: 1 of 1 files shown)


jshom:
Can you share an example where we have duplicate sections. img/screenshot or sql query result is good.
i.e. Pre & Post Fix
Thx

123joshuawu:
Pre fix
Screen Shot 2020-03-07 at 10 47 13 PM
Post fix
Screen Shot 2020-03-22 at 11 59 22 AM

PR #69, opened 2020-03-24
Frontend login

bobmayuze:
Added login button


src/yacs-web/src/App.vue
```diff
@@ -1,13 +1,22 @@
 <template>
   <div id="root">
-    <router-view></router-view>
```
123joshuawu: Uhh what happened to router-view lol

bobmayuze: idk, should I keep it or what

PR #72, opened 2020-03-24
Add vuex and update to scheduling

123joshuawu:
Currently `courses`, `selectedCourses`, and `Scheduler` are part of the global state and hence stored in the parent component `Main.vue`. This means any changes have to be bubbled up using events and any access have to be passed down via props. The alternative is to use a central store `vuex` to manage global state and provide a standard and centralized interface for accessing and changing the global state.

Side note:
Added nullish-coalescing operator and optional chaining to babel


123joshuawu:
I think that this branch has reached somewhat of a stable state and I have a clearer idea of where it is going. I want #64 to merge before this one but am looking for critique of the current state. 
Is the store useful for other functionality (Admin)? Need something else?
Is the organization of the code understandable? Too confusing? 
Is the store too dense? Should the code be separated into modules?
Is the design too hard to implement? Too verbose? Not verbose enough?
Idk what are your thoughts @CoreyBuckley @jshom @bobmayuze @marchdan

marchdan:
@123joshuawu Now that #64 has been merged, what is the status of this?

123joshuawu:
@marchdan I'll try to wrap this up today or tmrw. I'll leave the semester switching as it is and after this is merged, you can decide if you want to change it up.

marchdan:
@123joshuawu Cool. I think the semester switching should be good. But I’ll look at the new structure and decide then.

123joshuawu:
Closing this pr for now and will look to pick it up at a later time. Because it seems like there isn't an urgent need for this and introducing this correctly will require some restructuring, I'm putting this on the back burner. Most likely will see if I can pick it back up after this semester ends. And hopefully at that point this will be more useful to the project.

PR #75, opened 2020-03-27
Admin panel redesign

### LINKED ISSUE #85: Distribution — Admin Panel
Currently the admin panel is a list of modals for the various admin actions. Redesigning this while maintaining "/admin" as a hub for all possible admin actions would be nice.

CoreyBuckley:
Of the time I spent dedicated to this project over spring break, I mostly was prototyping an admin panel. As can be seen, we have multiple admin routes/actions now (csv, date binding, and seems selected semester will be one in the future) so this may or may not be helpful.

This was my beautiful sketch of my idea (I have no idea how to get rid of the buttons covering stuff up)

image

Because of #65 and the comment to move admin components to `/admin`, I've decided to open up this PR even though I'm not done and still have plans for it. I haven't had as much time to dedicate to it since spring break is over.

The design is inspired by the TFS dashboard where you could drag & drop these colorful and customizable cards that are associated with an admin action. I've also implemented a "global" loading bar. I've done this by specifying an interceptor/middleware on the axios object, so whenever it makes an API call, some state on the global Vue object (via a plugin) is modified to let `Admin.vue` know when to re-render. Vuex would be helpful for this, as I've been kind of tiptoeing around installing it, but with #72 in progress, I'm hopeful I can utilize it to store user settings and metrics on the axios API calls.

Note, the axios interceptor is on ONE instance of the axios object which is contained in AdminService.js. If there's some background operation that needs to make a network request, it's easy to make sure it's not tracked in the dashboard load bar.

Here's what the main screen of '/admin' looks like as of now:

image

Here's what it looks like to have admin pages open:

image

The tabs are lazy (except for dashboard), components are mounted when the tab is active. I would like to change this so when you click a tab it goes to '/admin/{component name}' but as of now it stays on '/admin'.

Here's what it looks like to make a (tracked) API call:

TrackNetworkReq

(That little delay you see before it "fully" finishes is the amount of time to wait in case another request happens, in which case the load process restarts. So that means it will load infinitely if you spam network requests? As of now, sure. Probably some client side error if you're spamming network requests tho)

Anyway, figured I should put this out here for now since it could be a solution to having everything at "/admin".

Closes #85


jshom:
going to close this until we decide to start the issue

PR #76, opened 2020-03-27
User System

liabenny:
Add user system backend APIs.

Functions:
- Create user
- Update user
- Delete user
- Get user profile
- Log in
- Log out
- Add user event


src/data/schema/user.sql
```diff
@@ -0,0 +1,63 @@
...
+	name TEXT,
+	email TEXT NOT NULL,
+	phone TEXT,
+	password TEXT,
+	major TEXT,
+	degree TEXT,
+	enable BOOLEAN DEFAULT TRUE
+);
+
+ALTER TABLE users owner TO postgres;
```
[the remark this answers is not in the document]
liabenny: Do you mean the 'enable' attribute? This one marks the user is in use, and it would be set false after deleting the user. In this case, we are able to collect user information cause we don't delete it.

jshom:
Appears to still be merge conflict here, can you clean this up ASAP, thx!

123joshuawu:
There seems to be some conflict regarding the return format of `database.execute`.
The existing code base would return `(ret, err)` while the new User code relied on `ret`.
I've opted to change all the User code to only consider `(ret, err)[0]` but for future purposes it would be better to handle database errors.

src/api/view/message.py
```diff
@@ -0,0 +1,11 @@
+from flask import jsonify
```
smgibson19, line 1: I see this import jsonify in a lot of the files, what from json is it importing?

123joshuawu: So jsonify is a function provided by Flask that transforms the RealDictRows returned by PostgreSQL into JSON.

PR #77, opened 2020-03-27
Finish dockerization on backend and postgres sql

bobmayuze:
Finish dockerization on backend and postgres sql


jshom:
Works with backend and db! Good stuff!
Lets create an nginx instance to serve front end and so that we can cache routes in the future for the api.

jshom:
Was a change to the return structure of get class full, needed to change test to accept this new return type. Tuple instead of value.

PR #81, opened 2020-03-29
ui tweaks, still needs more for beta release, deadlinks in footer

jshom:
We will be releasing beta soon, decided to add some design tweaks so we can read for v0
Screen Shot 2020-03-29 at 6 29 17 PM


123joshuawu:
Looks good to me, also just letting you know that `yacs.n/package-lock.json` got deleted

CoreyBuckley:
Footer looks good! Minor thing is I think it would be nice if you could move the "export" button to be above the schedule. For my screen size, I have to scroll down in order to see the button. Feel like it would make more sense to have be visible "above the fold".

CoreyBuckley:
kinda like this

image

or this

image

what you think?

123joshuawu:
Maybe we could move it with selectedcourses?

jshom:
@123joshuawu ye, I was thinking we could make the sub semester tabs just like the course search and selected courses. What you think?

Regarding the position of the export to ics, lets leave it as is. Once we have more export options I think the design choice will be easier to make.

PR #87, opened 2020-04-03
Fix subsemester course filter not showing up. Clean some of code.

CoreyBuckley:
Added default semester of summer 2020 in db build for now so subsemester filter is visible.


src/data/schema/admin_setting.sql
```diff
@@ -4,4 +4,4 @@ create table admin_settings(
 );
 
 INSERT INTO admin_settings (semester)
-VALUES (' ');
+VALUES ('SUMMER 2020');
```
123joshuawu, line 7: Should we opt for a specific value as the default value?

123joshuawu: Ok read pr comment resolved

jshom, line 7: We should not have a default that is RPI specific or time specific in table definitions, this must be extracted out to the get semester default. 
One solution could be to check if return value is Null from query, then select the first semester that is found from the list of classes.

PR #90, opened 2020-04-03
Undo

### LINKED ISSUE #88: Remove time specific default value
We should not have a default that is RPI specific or time specific in table definitions, this must be extracted out to the get semester default. 
One solution could be to check if return value is Null from query, then select the first semester that is found from the list of classes.

_Originally posted by @jshom in https://github.com/YACS-RCOS/yacs.n/pull/87_

CoreyBuckley:
> One solution could be to check if return value is Null from query, then select the first semester that is found from the list of classes.

As of right now, the solution for this could be messy since there is async code dependent on the currentSemester var value. Either wait for the courses network call to come back and then start the subsemester call, which will increase load time, or nest promises (less increase in load time), or nothing.

For right now, I've omitted this feature. You can see what it looked like in the past commit b72913e in Main.vue

Closes #88


src/web/src/pages/Main.vue
```diff
@@ -125,7 +125,11 @@ export default {
         this.currentSemester = semester[0].semester;
       });
     }
-    
+
+    getCourses().then(courses => {
+      this.courses.push(...courses)
```
jshom, line 130: why not do `this.courses = courses`?

PR #91, opened 2020-04-03
Minor UI fixes

marchdan:
Fixed some visual items on main page


src/web/src/components/Footer.vue
```diff
@@ -46,21 +49,42 @@
 
 import { getSemesters } from '@/services/YacsService';
 
+import { getSemester }  from '@/services/AdminService';
```
jshom, line 52: can we rename this to getDefaultSemester as that is what it represents, getSemester is ambiguous

src/web/src/components/Footer.vue
```diff
@@ -46,21 +49,42 @@
...
     created () {
-    getSemesters().then(({ data }) => {
-            this.semesterOptions.push(...data.map(s => ({text: s.semester, value: s.semester})));
-            });
+      if(this.$route.query.semester){
+        this.currentSemester = this.$route.query.semester;
+      }
+      else{
+        getSemester().then(semester => {
+          this.currentSemester = semester[0].semester;
```
jshom, line 70: Please change return from backend, having to chose first record of options then .semester is strange.
getDefualtSemeste should return default semester not rows of objects, i.e. "FALL 2020"  instead of [{"semester": "FALL 2020"}]

src/web/src/components/Footer.vue
```diff
@@ -46,21 +49,42 @@
...
     },
     created () {
-    getSemesters().then(({ data }) => {
-            this.semesterOptions.push(...data.map(s => ({text: s.semester, value: s.semester})));
-            });
+      if(this.$route.query.semester){
+        this.currentSemester = this.$route.query.semester;
+      }
+      else{
+        getSemester().then(semester => {
```
jshom, line 69: will require change here form name switch

src/web/src/components/Footer.vue
marchdan: Done

src/web/src/components/Footer.vue
marchdan: done

src/web/src/components/Footer.vue
marchdan: Yeet

PR #96, opened 2020-04-05
Fetch course metadata

### LINKED ISSUE #71: Distribution — RPI: Full Course Details
**Job**
As someone signing up for courses, I want to see the full description so I can avoid going to the catalog to know what the course entails.

**To-do:**
- [X] In the RPI parsing script, fetch details from the course catalog to add full course descriptions and full course names to the CSV.
- [X] Update the course upload functionality in Admin panel to allow for full course descriptions

This has already been implemented in old YACS in Ruby, we need to migrate it to Python.
It is located in pipelines/rpi/adapters/acalog-rpi

It appears that there is an API we can query for RPI catalog information at
http://rpi.apis.acalog.com

CoreyBuckley:
### Best Outcome

RPI standardizes their formatting for prerequisites and corequisites. Or better yet, separates them out into separate fields in the API response!

**Description**

This PR aims to grab additional course information that would be useful to display to the users and/or allow for additional features (e.g. feedback if a course doesn't meet a user's course history). 

Data comes from an undocumented acalog API; used current YACS ruby implementation to help develop a python interface for this API.

As of this post, the possible fields we can add to the CSV course information is:
```
ACALOG_COURSE_FIELDS = [
    "department",
    "level",
    "full_name",
    "description",
    "raw_precoreqs",
    "offer_frequency",
    "cross_listed",
    "graded",
    "credit_hours",
    "contact_lecture_lab_hours"
]
```
and of course, any combination/manipulation of these standard fields would be fair game for adding to the course info CSV. E.g. I've been using the field "short_name" which is just `${department}-${level}`

The response of the API is XML, this script turns it into JSON that is easier to pass around and read. The JSON is merged with the existing course information CSV (which comes from SIS).

**Features**

- Retrieve all course ids for a particular catalog from acalog API
- Retrieve all course info for a particular catalog from acalog API
- Retrieve the id for the current catalog from acalog API
- Little bit of threading so all requests for the course info are threaded. Implemented because it's not possible to get all course info in one GET request due to how the endpoint is structured. The XML to JSON (i.e. `fetch_catalog_course_info.py`) script doesn't take too long ~10s. 
- Control what fields are returned in the catalog course info JSON
- Ability to add custom fields (nothing fancy, just a place to put another `elif` and implement)
- Refactor `rpi-parse.py` to act more as a 'main' class. Parsing the SIS course (which @marchdan worked on I believe) info has been put into its own python module
- Tests for preco req parsing

**To-Do:**

- [X] Add frequency, prereq, coreq, description to schema for course tbl
- [X] Apply first normal form so that a course record doesn't have many prereqs. May need to create another table when introducing coreqs.
- [ ] Find a way to measure accuracy of pre/co req parsing??
- [X] Modify `get_course_info_full` to include prereqs & coreqs. 2 joins.
- [X] Get current catalog id

**Flow** (partially so I don't forget):

- Build the database schema
- Run `rpi-parse.py`
   - Will get all SIS & Acalog catalog course info, join, and output to CSV specified by an env var
- Run `courses.py` (or from admin UI the CSV import function)
   - Will parse the CSV file and distribute the data across tables accordingly. For each prereq, an entry in `course_prerequisite`. For each coreq, an entry in `course_corequisite`.

(Can also manually run `fetch_catalog_course_info.py` and generate the corresponding XML & JSON files)

**External dependencies introduced:**

- `lxml` for fast & efficient XML parsing. Very important as the generated XML is ~28MB or so
- `pytest-check` to further control the output of pytest. Allows 'fail-through'; can see rest of the 'assert' statements even if previous one failed.
- `regex` allows for branch reset regex. Makes it easier to manage many capture groups. [relevant SO post](https://stackoverflow.com/questions/44460642/python-regex-duplicate-names-in-named-groups/44463324#44463324)

closes #71


CoreyBuckley:
So I don't forget, CodeFactor has a problem with lxml.parse because it's potentially vulnerable to XML attacks.

This is not a problem. The only source of XML we receive is from RPI's endpoint -- so any security is directly tied to that.

The parsing xml script does not allow for user inputted XML nor is the plan for a user to upload input via the UI.

For more information: [lxml security](https://lxml.de/FAQ.html#is-lxml-vulnerable-to-xml-bombs)

CoreyBuckley:
This feature is pretty much finished with the one exception being accuracy of course corequisites and prerequisites. 

Due to the large number and unpredictability of possible course precoreq text, one of the following would need to happen:
    - the regex will have to be continuously updated
    - a new approach is taken that isn't regex 
    - the maintainers of the acalog api standardize their XML response so it's easy to parse prereqs and coreqs.

123joshuawu:
much impress, much wow!! I think training a model for parsing the response might not be bad idea. I was kind of joking when I first brought it up but it could be viable? Maybe could start out by comparing it against the regex results? @bobmayuze 
Otherwise overall it looks excellent, 5 michelin star

On other note just curious on your thoughts for choosing first normal form over second normal form for pre/corequisites

CoreyBuckley:
@123joshuawu Thanks 😃 

The added pre/co req tables are already in the strongest normalization possible, as all columns are the key.

By first normal form, I just meant that a course record didn't have an array or string of "ECSE-2210, ADMN-1200, ...". 

Yuze showed Jacob and I this amazon [NLP](https://console.aws.amazon.com/comprehend/v2/home?ad=c&cp=bn&p=cmp&region=us-east-1#home) but sending an API call for each course is a lot of overhead. Would be cool if we could train our own model to do it though, haha.

PR #98, opened 2020-04-08
Ops — Help us review/build fast

jshom:
**On a PR that tries to merge a branch into master**
1. Spin up a new server with that branch's code
2. Share's link through comment


CoreyBuckley:
Added both summer and fall 2020 csvs, set default semester for fall and summer, changed semester part name mapping, exported to schedule.

All seems to be working as usual. Don't know a lot of devops so can't really comment on the code atm. Gets a pass from me.

PR #107, opened 2020-04-10
Added links to footer

marchdan:
This fills in the previously empty links in the yacs.n footer


jshom:
Issue stated cookie policy, but after discussion, we decided to defer until we have the names and uses of those cookies to accurately define our cookie policy.

PR #109, opened 2020-04-17
Save selected courses

marchdan:
Functionality to save the courses the user has currently selected. 

**Testing Procedure:**

1. Open a terminal and curl a user account:
- `curl -d '{"name":"John Doe", "email":"yacs@rpi.edu", "phone":"1234567890", "major":"CS", "password":"123456", "degree":"Undergraduate"}' -X POST _url_/api/user -H "Content-Type: application/json" `
2. Login to the page with your email and password
3. navigate to _url_/admin
4. Upload the Fall, Spring, and Summer 2020 CSV files.
5. Navigate back to the home page and add courses for each semester. 
- Semester can be changed in the footer
- Make sure to select courses for all there subsemesters of the Summer semester
6. Ensure that when you return to a semester, or refresh the page, all of your previously selected courses still appear in both selected courses, and on your schedule.
7. Log out, refresh the page (all selected courses should be gone), then log back in.
8. Ensure all of your selections are still present after this proccess.
9. Celebrate because this testing process is long


scripts/dev-start-legacy.sh
```diff
@@ -12,7 +12,7 @@ echo "Starting @ $(date)" > $PATH_ROOT/logs/dev.txt
 
 # run flask
 cd src/api/
-flask run --reload >> $PATH_ROOT/logs/api.txt 2>&1 & # run in bg and port logs
+DB_USER=marchdan DB_PASS=" "  flask run --reload >> $PATH_ROOT/logs/api.txt 2>&1 & # run in bg and port logs
```
CoreyBuckley, line 15: I'm guessing that DB_USER shouldn't probably be your own account?

marchdan: Crap, I thought I unstated that file XD Will fix on next push

src/api/db/student_course_selection.py
```diff
@@ -0,0 +1,37 @@
+class student_course_selection:
+	def __init__(self, db_conn):
+		self.db_conn = db_conn
+
+	def add_selection(self, cid, sem, uid):
+		sql = 	"""
+				INSERT INTO
+					student_course_selection (user_id, semester, course_id)
+				VALUES
+					(%d, %s, %s)
```
CoreyBuckley, line 10: I believe the psycopg2 page says that you should always use `%s` for the placeholder. It's not like a C format flag where it corresponds to some type.

marchdan: Thank you, will fix for next push.

src/api/db/student_course_selection.py
```diff
@@ -0,0 +1,72 @@
...
+						student_course_selection (user_id, semester, course_id)
+					VALUES
+						(%s, %s, %s)
+					"""
+			resp, error = self.db_conn.execute(sql, [uid, sem, cid], False)
+		
+		except Exception as e:
+			return (False, e)
+
+		if error is not None:
```
CoreyBuckley, line 18: I don't think `error` is available at this part of the code.

src/api/db/student_course_selection.py
```diff
@@ -0,0 +1,72 @@
...
+						user_id = %s AND
+						semester = %s AND
+						course_id = %s
+					"""
+			resp, error = self.db_conn.execute(sql, [uid, sem, cid], False)
+
+		except Exception as e:
+			return (False, e)
+
+		if error is not None:
```
CoreyBuckley, line 38: same thing

src/api/db/student_course_selection.py
```diff
@@ -0,0 +1,72 @@
...
+						user_id = %s
+					order by
+						course_id asc
+					"""
+			courses, error = self.db_conn.execute(sql, [uid], True)
+
+		except Exception as e:
+			return (False, e)
+
+		if error is not None:
```
CoreyBuckley, line 64: same thing

src/data/schema/10_student_course_selection.sql
```diff
@@ -0,0 +1,21 @@
...
+
+INSERT INTO student_course_selection (user_id, semester, course_id) 
+VALUES (
+    (SELECT DISTINCT
+        user_id
+    FROM
+        user_account), 
+    null,
+    null
+);
```
CoreyBuckley, line 20: ummm, you can't do this. You're mixing a relation and a singular value with this syntax. You either provide a relation or a series of values corresponding to the table's columns.

So change it to 

```
INSERT INTO student_course_selection (user_id, semester, course_id) 
(
    SELECT DISTINCT
        user_id,
        NULL,
        NULL
    FROM
        user_account
);

File: src/web/src/pages/Main.vue [hunk 7]
```diff
@@ -1,6 +1,6 @@
 <template>
   <div>
-    <Header class="mb-3" :currentSemester="currentSemester"></Header>
```
CoreyBuckley, line 3: is there a reason for removing this?

src/web/src/pages/Main.vue
```diff
@@ -13,7 +13,6 @@
                     @removeCourse="removeCourse"
                     :courses="courses"
                     class="w-100"
-                    :selectedSemester="currentSemester"
```
CoreyBuckley, line 16: why remove this?

src/api/db/student_course_selection.py
```diff
@@ -0,0 +1,72 @@
+class student_course_selection:
+	def __init__(self, db_conn):
+		self.db_conn = db_conn
+
+	def add_selection(self, cid, sem, uid):
+		try:
```
CoreyBuckley, line 6: You shouldn't have to wrap these queries in a `try` block. The `db_conn` is already a wrapper around the psycopg2 driver which will do the actual DB query call in a `try` block and return the corresponding error on exception like you have done here.

src/api/db/student_course_selection.py
marchdan: Moved

src/web/src/pages/Main.vue
```diff
@@ -156,6 +183,21 @@ export default {
       // This must be vm.set since we're adding a property onto an object
       this.$set(this.selectedCourses, course.id, course);
       this.scheduler.addCourse(course);
+
+      if(this.userID){
+        const info = {'cid':course.name, 'semester':this.currentSemester, 'uid':this.userID};
+
+        console.log(info);
+        addStudentCourse(info)
```
CoreyBuckley, line 191: I'm not sure if security is required for this PR, but it would be very easy to `spoof` my UID and add courses to other users due to a lack of checking. After #111 is merged in, it won't be too hard to secure this route.

src/api/db/student_course_selection.py
marchdan: Moved

src/api/db/student_course_selection.py
marchdan: Moved

src/data/schema/10_student_course_selection.sql
marchdan: Will do, wasn't sure how this would work

src/web/src/pages/Main.vue
marchdan: Has been put back

src/web/src/pages/Main.vue
marchdan: Has been undone

src/data/schema/10_student_course_selection.sql
CoreyBuckley: For consistency, we decided this block should be removed as there is no reason for initialization and NON-NULL constraint as part of PK will make this impossible anyway.

src/web/src/pages/Main.vue
marchdan: No one gave me a way of getting the uid, no matter how many times I asked, so I just put it in cookies. Feel free to fix this, but I can't do better with the information I have been given...

CoreyBuckley: That is fine, no problem.

src/api/app.py
```diff
@@ -205,6 +207,24 @@ def log_out():
...
+    info = request.get_json()
+    resp, error = course_select.add_selection(info['name'], info['semester'], info['uid'], info['cid'])
+    return Response(status=200) if not error else Response(error, status=500)
+
+
+@app.route('/api/removecourse', methods=['POST'])
+def remove_student_course():
+    info = request.get_json()
+    resp, error = course_select.remove_selection(info['name'], info['semester'], info['uid'], info['cid'])
+    return Response(status=200) if not error else Response(error, status=500)
```
CoreyBuckley, line 221: reason for not doing methods=['DELETE']? 🤔

src/api/app.py
```diff
@@ -205,6 +207,24 @@ def log_out():
...
+
+@app.route('/api/getcourses', methods=['POST'])
```
CoreyBuckley, line 223: reason for not doing methods=['GET']?

src/api/db/student_course_selection.py
```diff
@@ -0,0 +1,61 @@
...
+				VALUES
+					(%s, %s, %s, %s)
+				"""
+		resp, error = self.db_conn.execute(sql, [uid, sem, name, cid], False)
+		return (True, None) if not error else (False, error)
+
+		
+
+	def remove_selection(self, name, sem, uid, cid):
+		if cid == '-1':
```
CoreyBuckley, line 18: is `cid` being -1 a way for the dev to specify to delete all courses? Feels a little obscure. Possible to add documentation or make cid a keyword argument?

src/web/src/pages/Main.vue
```diff
@@ -117,18 +117,59 @@ export default {
...
   },
   async created() {
     this.currentSemester = this.$route.query.semester ? this.$route.query.semester : await getDefaultSemester();
-    this.updateDataOnNewSemester();
+    
+    console.log(`Semester: ${this.currentSemester}`);
+    
+    await this.updateDataOnNewSemester();
+
+    this.userID = this.$cookies.get("userID");
```
CoreyBuckley, line 134: It would be nice if you could move the fetching of user courses into its own function and then call it in created so we don't have to reload the whole page everytime a user changes semester (which also incurs the cost of Main.vue having to be re-mounted.)

src/api/app.py
marchdan: Can you pass an argument to 'DELETE'?

src/api/app.py
marchdan: Not easy to pass an argument to 'GET', if there is a way I wasn't using, lmk

src/api/db/student_course_selection.py
marchdan: Can't use none if it is part of primary key. I would love to use a better value if you have any ideas

src/web/src/pages/Main.vue
marchdan: Including it in promises.all will ensure the promises return before this is called?

src/api/db/student_course_selection.py
marchdan: Also, -1 is to specify it as the course object, instead of a specific section, it's not a specifier to delete all. It's just that if the course object is being deleted, all associated sections also have to be deleted

src/api/app.py
CoreyBuckley: Yes.

https://github.com/YACS-RCOS/yacs.n/blob/clean_semester_upload_and_preview/src/web/src/services/UserService.js#L8-L15

jshom:
@marchdan 
Can you provide a script to test?
Lets make it easy for contributors to check functionality, thanks!

src/api/app.py
CoreyBuckley: if all you're passing to the GET is 
```
const info = {'uid': this.userID};
```
then you can just include that as 
```
uid => client.get('/getcourses', {
    params: {
        "uid": uid
    }
})
```
and update your flask handler to use `request.args['uid']` instead

src/web/src/pages/Main.vue
CoreyBuckley: Yes, if you include it in the then() callback.

src/api/db/student_course_selection.py
CoreyBuckley: I just mean you would change the condition `if cid == '-1'` to `if cid is None`. So that cid is an optional parameter.

src/api/app.py
marchdan: Done

src/api/app.py
marchdan: Done

src/api/db/student_course_selection.py
marchdan: Done

src/web/src/pages/Main.vue
marchdan: Can't await in a non-async function, and can't declare method async

marchdan: Fixed. Please review upon push

123joshuawu:
@marchdan this should fix the error
```python
def test_model_factory(model):
    class TestModel(model):
        def __init__(self, db):
	    self.db = db
    return TestModel

@pytest.fixture(scope="session")
def test_user(db_conn):
    return test_model_factory(User)(db_conn)
```

123joshuawu:
ugh ok we need to move the unit tests

marchdan:
We will move them with the refactor

marchdan:
@jshom @123joshuawu @CoreyBuckley Who's gonna review?

scripts/dev-start-legacy.sh
jshom: quick fix here, can you remove your specific ENV vars

tests/api/db/test_saving.py
```diff
@@ -0,0 +1,45 @@
+def test_save_student_courses(save_semester, test_user):
+	args = {"name":"Barb", "email":"user1@rpi.edu", "phone":"1234567890", "major":"CS", "password":"123456", "degree":"Undergraduate"}
```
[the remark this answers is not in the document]
marchdan: Will change, but be aware you're commenting on an old file that is currently commented out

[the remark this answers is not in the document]
jshom: oops forgot to submit review earlier

src/api/app.py
```diff
@@ -205,6 +207,24 @@ def log_out():
 def add_user_event():
     return event_controller.add_event(json.loads(request.data))
 
+@app.route('/api/addcourse', methods=['POST'])
```
CoreyBuckley, line 210: conforming with the recent meeting, would be nice if you could change the endpoint names to be `/api/course` and then just have the method vary (e.g. POST would be equivalent to your old '/api/addcourse', etc.)

src/api/app.py
```diff
@@ -205,6 +207,24 @@ def log_out():
...
+def add_student_course():
+    info = request.get_json()
+    resp, error = course_select.add_selection(info['name'], info['semester'], info['uid'], info['cid'])
+    return Response(status=200) if not error else Response(error, status=500)
+
+
+@app.route('/api/removecourse', methods=['DELETE'])
```
CoreyBuckley, line 217: same here

marchdan: Done

src/api/app.py
marchdan: Done

CoreyBuckley:
I have tested according to your testing procedure. All was working for me. Very cool!

I encourage you to remove your debug `console.log`s before merge however.

marchdan:
> I have tested according to your procedure and all was working. Very cool. I encourage you to remove your debug `console.log`s before merge though.

The ones I intended to remove have been. If there are other ones which need be removed I can do so

src/web/src/pages/Main.vue
```diff
@@ -149,12 +190,42 @@ export default {
...
+
+          addStudentCourse(info)
+            .then(response => {
+              console.log(response);
+            })
+            .catch(error => {
+              console.log(error.response);
+            });
+
+          console.log(`Saved section ${section.crn}`);
```
CoreyBuckley, line 225: This console.log is not always true. You should put it in the .then()

src/web/src/pages/Main.vue
```diff
@@ -165,13 +236,44 @@ export default {
...
+        
+        removeStudentCourse(info)
+          .then(response => {
+            console.log(response);
+          })
+          .catch(error => {
+            console.log(error.response);
+          });
+
+        console.log(`Unsaved ${course.name}!`);
```
CoreyBuckley, line 251: same here

PR #111, opened 2020-04-19
Add hidden semesters & remove all data then reinsert on CSV upload

### LINKED ISSUE #105: Distribution — Semester Upload
**Job**
When an admin uploads a semester, the admin uploads a full semester and therefore wants to update all data.

**Details:**
- Allow an option to hide on csv upload, so that an admin can view a semester's data without publishing to everyone just yet.
- Allow show/hide of semester in admin panel
  - Rename admin module (Rename Semesters Parts) to (Edit Semesters)
  - Insert this show/hide option as checkmark next to the semester name i.e. FALL 2020 [ X ]

**Context**
Right now we have append only upload, does not factor in courses being removed from a semester or times changing. This way, we can upload/update to accurate information.

Also, this will then allow us to remotely send files and have auto updates by parsing and shooting csv's at the api

### LINKED ISSUE #110: Bug — Fix filtering of semester
**Describe the bug**
Currently, the classinfo.py class is returning ALL courses in the db. This is making our UI slower as it forces filtering on front end and takes longer to load the JSON request.

Instead of filtering on the frontend, parametrize the classinfo.py class and edit API calls to fetch classes by using the format /api/class/FALL-2020.

image

CoreyBuckley:
## THIS PR IS READY FOR REVIEW

Code is a little messy atm, focusing on the feature working rn.

Fixed user account logout, added user tables & semester_info to build script. Non-public semesters are hidden to all users except admins in the footer UI.

- [x] Need to wipe courses & course_session on CSV upload prior to insertion
- [X] Need to protect routes for semesters that are specified to be non-public.
image

- [X] Fix client-side filtering of all db courses by parameterizing query by semester
image
- [X] Should reconsider what the 'default' semester is. Choose the first public one?

- [X] Changed footer anchor tags to btns that will update the current page. This means an API request for each time you change semesters and you will see some delay. Am not sure how big of a deal this will be. To make this prettier, I believe there is a way to 'fake' changing the page in a SPA app (so the url changes) but the state of the SPA changes. Wasn't too concerned with look for this PR however.
image

**NOTE**

I am using [flask sessions](https://flask.palletsprojects.com/en/1.1.x/quickstart/?highlight=middleware#sessions) instead of the `user_session` tbl. Flask session makes it easier for me to persist the logged in user data rather than having to query the session tbl for a uid then query users with that uid. If this is a problem, it shouldn't be too hard to refactor.

In this PR, the sessionID that is part of user_sessions table serves only the purpose of recording log ins and log outs.

**How to test**:

1. Rebuild DB via `src/db/build.sh`
2. Create an admin user via psql
```
INSERT INTO user_account VALUES (
    1,
    'admin',
    'admin@rpi.edu',
    '2904092',
    'f52fbd32b2b3b86ff88ef6c490628285f482af15ddcb29541f94bcf526a3f6c7',
    'cs',
    'under',
    1::boolean,
    1::boolean,
    1::boolean
);
```
3. Go to `/admin` and upload two CSVs where one of them is non-public, can do this at upload stage or in `Edit Semesters` action on admin panel.
4. Go to schedule page and look at footer and make sure the non-public link isn't there.
5. Go to the non-public link via URL and you should get an HTTP 401 unauthorized error.
6. Login with user "admin" and password "hunter2"
7. The non-public link in the footer should be there.
8. The course data should be loaded for that semester.
9. Press logout and a timestamp should be added to `user_session` tbl and your session cookies will be cleared.

**IMPROVEMENTS THAT COULD BE MADE**
- Separate page for unauthorized request? Or some indication beyond dev tools?
- Indication that currently looking at a non-public semester on the schedule UI.
- Make Edit Semesters component more robust; separate out validation logic, tweak organization.

Closes #105
Closes #110


src/web/src/components/EditSemesterDateNameBinding.vue
```diff
@@ -4,7 +4,16 @@
         ref="editSemesterForm"
         @submit.prevent="onSubmit"
     >
+        <input type="hidden" name="semesterTitle" :value="semesterTitle" />
```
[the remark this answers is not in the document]
CoreyBuckley: ok, ty 🤖

src/web/src/pages/Main.vue
```diff
@@ -118,20 +118,11 @@ export default {
...
-    } else {
-      getDefaultSemester().then(semester => {
-        this.currentSemester = semester;
-      });
-    }
+  async created() {
+    this.currentSemester = this.$route.query.semester ? this.$route.query.semester : await getDefaultSemester();
 
-    getCourses().then(courses => {
+    Promise.all([getCourses(this.currentSemester), getSubSemesters(this.currentSemester)]).then(([courses, subsemesters]) => {
```
123joshuawu, line 124: If the function is already `async` why opt for `Promise.all` over separate `await`s?

src/web/src/services/YacsService.js
```diff
@@ -36,8 +36,8 @@ const _getCourseIdentifier = courseObj => {
  * Returns a list of all courses
  * @returns {Promise<Course[]>}
  */
-export const getCourses = () =>
-  client.get('/class').then(({ data }) => {
+export const getCourses = (semester) =>
+  client.get(`/class?semester=${semester}`).then(({ data }) => {
```
123joshuawu, line 40: nitpick but can provide url paramters to `axios` through url parameters, think its like `get(..., {params: {semester})` or something

src/web/src/pages/Main.vue
```diff
@@ -118,20 +118,11 @@ export default {
...
   },
-  created() {
-    if (this.$route.query.semester) {
-      this.currentSemester = this.$route.query.semester;
-    } else {
-      getDefaultSemester().then(semester => {
-        this.currentSemester = semester;
-      });
-    }
+  async created() {
```
123joshuawu, line 121: since we're very much operating as a SPA and we're passing currentSemester via props, can probably change footer to emit changes to `currentSemester` and avoid reloading the whole page

src/api/app.py
```diff
@@ -134,11 +184,18 @@ def update_user_info():
 
 @app.route('/api/session', methods=['POST'])
 def log_in():
-    return session_controller.add_session(request.json)
+    session_res = session_controller.add_session(request.json).json
```
123joshuawu, line 187: Me gusta using flask sessions, later we can look to deprecate the session controller

src/web/src/pages/Main.vue
CoreyBuckley: await will halt the execution of the main thread until that function has finished, no? Promise.all allows these actions to occur async at the same time

src/api/app.py
```diff
@@ -25,8 +29,16 @@
 courses = Courses.Courses(db_conn)
 date_range_map = DateMapping.semester_date_mapping(db_conn)
 admin_info = AdminInfo.Admin(db_conn)
+semester_info = SemesterInfo.semester_info(db_conn)
+users = UserModel.User()
 
 app = Flask(__name__)
+app.secret_key = "someRandomStuffForNow"
```
jshom, line 36: Please make this an env var. We should not have exposed keys/secrets.
- [ ] Fetch with `os.environ`
- [ ] Add env var in dev & prod docker-compose with default like `${SECRET:-default_key}`, should be a few examples in docker-compose.production.yml

src/web/src/pages/Main.vue
CoreyBuckley: that makes sense, I'll look into it.

src/web/src/services/YacsService.js
CoreyBuckley: ok, will do

src/web/src/pages/Main.vue
123joshuawu: True, I guess its a readability vs optimization thing

src/web/src/pages/Main.vue
CoreyBuckley: Ok, I've done it the way you described. Because course fetching is parameterized by currentSemester each time you change semester a new API call occurs. Wonder if ppl will change semester enough for this to be a problem?? Hopefully not.

src/api/app.py
CoreyBuckley: :smile:

src/api/app.py
CoreyBuckley: Done

PR #131, opened 2020-04-24
Fixing UX bugs from #126

marchdan:
**Issue**
closes #126 

**Test Procedure**

Will be updated as features are fixed:

> 1. Load public semesters through admin page
> 2. On main page, look at footer and see that the other semesters are black, bolded text
> 3. Click on the title of a course in Course search, and make sure it gets added to Selected courses
> 4. Go to selected courses and click on the title of the course, the sections should collapse and the chevron should switch from pointing up to pointing down.
> 5. Remove all courses from selected courses, a warning should appear telling you to add courses
> 6. Leave a comment with thoughts about the warning and how it should be improved. 

**Photos**
`Before:`
other-semesters-before
`After:`
other-semesters

`Before:`
no-course-before
`After:`
no-course-after

PR #138, opened 2020-04-24
Add course information to CourseList

### LINKED ISSUE #134: Course Search — Pre/Coreq and description info
**Job**
When I'm searching for issues, I want to be able to see the course description and pre/co-reqs

Show a (info) button that drops down extended course info

**Context**
This data is in the api all response in /class

123joshuawu:
**Issue**

Closes #134 

N/A

**Test Modifications**
N/A

**Test Procedure**

Click on information icon to see course information including if available: coreqs, prereqs, and description

**Photos**

Screen Shot 2020-04-24 at 4 35 23 PM
Screen Shot 2020-04-25 at 12 34 37 PM
Screen Shot 2020-04-25 at 12 34 11 PM
Screen Shot 2020-04-25 at 12 33 54 PM
Screen Shot 2020-04-25 at 12 33 23 PM
Screen Shot 2020-04-25 at 12 33 12 PM

**Additional Info**
Refactor CourseListing to use slots. CourseList now overrides default slots to show course information

### DIFF
--- a/src/web/src/components/CourseList.vue
+++ b/src/web/src/components/CourseList.vue
@@ -49,21 +49,21 @@
             </button>
           </template>
           <template
-            #collapseContent="{ course: {corequisites, prerequisites, raw_precoreqs, description} }"
+            #collapseContent="{ course: {corequisites, prerequisites, raw_precoreqs, frequency, description} }"
           >
-            <span v-if="prerequisites">
-              Prereqs: {{prerequisites.join(", ")}}
-              <br />
-            </span>
-            <span v-if="corequisites">
-              Coreqs: {{corequisites.join(", ")}}
-              <br />
-            </span>
-            <span v-if="!prerequisites && !corequisites && raw_precoreqs">
-              {{raw_precoreqs}}
-              <br />
-            </span>
-            <span v-if="description">{{description}}</span>
+            <div class="ml-3">
+              <span v-if="frequency">
+                Offered: {{frequency}}
+                <br />
+                <br />
+              </span>
+              <span>{{generateRequirementsText(prerequisites, corequisites, raw_precoreqs)}}</span>
+              <span v-if="description">
+                <br />
+                <br />
+                {{description}}
+              </span>
+            </div>
           </template>
         </CourseListing>
       </b-list-group-item>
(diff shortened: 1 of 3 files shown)


123joshuawu:
will probably want to get #131 merged in before this one

src/web/src/components/CourseList.vue
```diff
@@ -28,12 +28,33 @@
...
+          <template #toggleCollapseButton="{ course, toggleCollapse }">
+            <button
+              class="btn"
+              @click="toggleCollapse()"
+              :disabled="!course.corequisites && !course.prerequisites"
+            >
+              <font-awesome-icon :icon="faInfoCircle" />
+            </button>
+          </template>
+          <template #collapseContent="{ course: {corequisites, prerequisites, description} }">
```
CoreyBuckley, line 46: What does the hash sign do in this case?

src/web/src/components/CourseList.vue
```diff
@@ -28,12 +28,33 @@
...
         :key="course.id"
         :class="{'bg-light': course.selected}"
       >
-        <CourseListing :course="course" :actions="{add:true}" v-on="$listeners" />
+        <CourseListing :course="course" lazyLoadCollapse v-on="$listeners">
+          <template #toggleCollapseButton="{ course, toggleCollapse }">
+            <button
+              class="btn"
+              @click="toggleCollapse()"
+              :disabled="!course.corequisites && !course.prerequisites"
```
CoreyBuckley, line 41: && !course.raw_precoreqs

src/web/src/components/CourseList.vue
```diff
@@ -28,12 +28,33 @@
...
+            >
+              <font-awesome-icon :icon="faInfoCircle" />
+            </button>
+          </template>
+          <template #collapseContent="{ course: {corequisites, prerequisites, description} }">
+            <span v-if="corequisites">
```
CoreyBuckley, line 47: i'm not sure what the condition should be, but show raw_precoreqs if the parsed out prereqs is empty?

CoreyBuckley:
Am getting 502 on temp server ☹️ 
 image

src/web/src/components/CourseList.vue
123joshuawu: I believe its a shorthand for named slots so the alternative is `v-slot:name=`

src/web/src/components/CourseList.vue
123joshuawu: yeah sounds good to me

src/web/src/components/CourseList.vue
123joshuawu: claro que si senor

CoreyBuckley:
The clickable space for adding a class seems to be obscuring the info btn.  Whenever I try to click the info btn, it just adds the course instead.

123joshuawu:
@CoreyBuckley just tried the temp deploy and seems to be working there
not sure why that would happen because i think events usually propagate from the child up

CoreyBuckley:
I was being dumb. Was clicking it on a class where it was disabled.

_Therefore_, I propose you get rid of the info button in place of it being disabled. Just leads to confusion.

jshom:
> I was being dumb. Was clicking it on a class where it was disabled.
> 
> _Therefore_, I propose you get rid of the info button in place of it being disabled. Just leads to confusion.

Agree

jshom:
Screen Shot 2020-04-24 at 9 29 21 PM

^ Looks a bit bland, lets add some spice

- Lets use the full_title instead of title
- Lets maybe use language of Require completion of CSCI-2500, 2300 instead of pre-req: x,y,z.
- Also, lets collect the additional info in a box of some sort, maybe block quote or some other way of grouping the additional info visually

123joshuawu:
Sounds good, lemme put those in

PR #140, opened 2020-04-25
Add back button functionality to course listing

### LINKED ISSUE #139: Bug — + Btn doesn't work on course list
**Describe the bug**
As of master 6a5095ebbb6ed862407d9ce57cb757c858c16a47, clicking the '+' button on a class listing does not add a class. (I think it's triggering the event twice, so it adds the class and then removes it)

**To Reproduce**
1. Go to Schedule
2. Click on '+' for a class listing 
3. Observe

**Expected behavior**
Button should add class

**Screenshots**

Kapture 2020-04-24 at 21 41 47

**Desktop (please complete the following information):**
 - OS: Win 10 & Mac OS

marchdan:
**Issue**
closes #139 

**Test Procedure**

> 1. Load in a semester of courses
> 2. Click on the plus to add a course, and make sure it adds
> 3. Go to the Selected courses tab
> 4. Click on the arrow to expand/collapse the section list, make sure it works

### DIFF
--- a/src/web/src/components/CourseListing.vue
+++ b/src/web/src/components/CourseListing.vue
@@ -9,13 +9,9 @@
       </div>
       <div>
         <button
-          v-if="actions.remove"
+          v-if="actions.add || actions.remove"
           class="btn"
-          @click="toggleCourse(course)"
-        >
-        <button
-          v-if="actions.add"
-          class="btn"
+          @click="actions.remove ? toggleCourse(course):null"
         >
           <font-awesome-icon v-if="course.selected" :icon="faTimes" />
           <font-awesome-icon v-else :icon="faPlus" />
--- a/src/web/src/components/CourseListing.vue
+++ b/src/web/src/components/CourseListing.vue
@@ -11,7 +11,6 @@
         <button
           v-if="actions.add || actions.remove"
           class="btn"
-          @click="toggleCourse(course)"
         >
           <font-awesome-icon v-if="course.selected" :icon="faTimes" />
           <font-awesome-icon v-else :icon="faPlus" />
@@ -19,7 +18,6 @@
         <button
           v-if="actions.collapse"
           class="btn"
-          @click="toggleShowSection()"
           :disabled="!course.sections.length"
         >
           <font-awesome-icon v-if="!this.showSections" :icon="faChevronDown" />
--- a/src/web/src/components/CourseListing.vue
+++ b/src/web/src/components/CourseListing.vue
@@ -9,7 +9,12 @@
       </div>
       <div>
         <button
-          v-if="actions.add || actions.remove"
+          v-if="actions.remove"
+          class="btn"
+          @click="toggleCourse(course)"
+        >
+        <button
+          v-if="actions.add"
           class="btn"
         >
           <font-awesome-icon v-if="course.selected" :icon="faTimes" />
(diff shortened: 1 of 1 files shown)


src/web/src/components/CourseListing.vue
```diff
@@ -11,15 +11,14 @@
...
-          @click="toggleCourse(course)"
+          @click="actions.remove ? toggleCourse(course):null"
         >
           <font-awesome-icon v-if="course.selected" :icon="faTimes" />
           <font-awesome-icon v-else :icon="faPlus" />
         </button>
         <button
           v-if="actions.collapse"
           class="btn"
-          @click="toggleShowSection()"
```
123joshuawu, line 22: er where did this go

src/web/src/components/CourseListing.vue
```diff
@@ -11,15 +11,14 @@
         <button
           v-if="actions.add || actions.remove"
           class="btn"
-          @click="toggleCourse(course)"
+          @click="actions.remove ? toggleCourse(course):null"
```
123joshuawu, line 14: this should be breaking the course list 
it should be sufficient to change `@click` to `@click.stop` to prevent the click event from propagating to the parent div

marchdan: This click event needs to happen if it's a sections listing

src/web/src/components/CourseListing.vue
marchdan: It happens when you click anywhere in the box, having this there will double click the event

src/web/src/components/CourseListing.vue
marchdan: Why would it be breaking the course list? If action.remove, the button toggles the course, else it does nothing

src/web/src/components/CourseListing.vue
123joshuawu: yes, this works right now because we aren't using the collapse element in `CourseList`. but if we added added `actions.collapse` to `CourseList` it wouldn't work.
if you add `@click.stop`, when you click on the button it won't pass the click to the box

src/web/src/components/CourseListing.vue
123joshuawu: ok my bad, it doesn't break because both `SelectedCourses` and `CourseList` provide both `actions.add` and `actions.remove`. In this case, if the `CourseListing` was only provided `actions.add`, then the add wouldn't work.

src/web/src/components/CourseListing.vue
marchdan: So the beginning would be click.stop, then add the click at the bottom?

src/web/src/components/CourseListing.vue
marchdan: Hmm? Course list provides .add & Selected courses provides .remove. That's what I am checking for here

src/web/src/components/CourseListing.vue
marchdan: If we added a drop down here, we'd have to decide what the default behavior of clicking the box would do, which would require a logic change here anyway

123joshuawu: so `@click.stop="toggleShowSection()"` and same for the other click

marchdan: So it stops the other click and does the new one?

src/web/src/components/CourseListing.vue
123joshuawu: Nevermind I'm wrong you're correct. but i'd much prefer if you used `@click.stop` instead

PR #190, opened 2020-08-07
Majors have classes and departments have majors

### LINKED ISSUE #168: Feature Request — Display the classes corresponding to each major
**Job**
Currently when you click on a major in the course explorer placeholder text is shown. This should be replaced with a list of the courses that are in that major.

I think what will work is reusing the course list component in order to accomplish this. Then the classes can split based on major and those can be passed into the course list. This also has the benefit of keeping a consistent style

**Context**
adds to #156

### LINKED ISSUE #185: Bug — Course Explorer isn't organized by department
**Describe the bug**
Each department should have its own list of majors, and as of right now they are organized alphabetically.

**To Reproduce**
1. Click on Explore link on the header
2. Look at the headings of the department lists

**Expected behavior**
Each "school" should be a header with all corresponding majors underneath

**Additional context**
Not really a "bug" per say but ya know

IkeLyons:
**Issue**
closes #185 
closes #168 

N/A

**Test Modifications**
N/A

**Test Procedure**
1) Click on Explore link
2) Explore classes by clicking on things

**Photos**
Before: a bunch of TODO messages....

After:
image

**Additional Info**
N/A


jshom:
With explore we want to show more info up front. Maybe we route to new page with full screen and expanded course details for that department. Would make it easy for people to share all available courses for a major (for their school)

What do you think?

src/web/src/components/DepartmentList.vue
```diff
@@ -17,7 +17,41 @@
             role="tabpanel"
           >
             <b-card-body>
-              <b-card-text>TODO: Insert Major's Classes</b-card-text>
+              <div id="scroll-box">
+                <recycle-scroller
```
123joshuawu, line 21: I'd suggest using `<list>` and `<listitem>` here instead of `<recycle-scroller>`

src/web/src/pages/CourseExplorer.vue
```diff
@@ -55,12 +103,37 @@ export default {
...
+      schoolsMajorDict: {},
       ready: false,
+
+      /*
+       *Used to define the order that the schools are placed into the two main
+       *columns. This could be implemented generically, but making the columns
+       *more or less "even" is NP-Hard so its more convenient to just define
+       *manually
+       */
+      schoolOrder: [
```
123joshuawu, line 115: what if you sort the schools by the number of courses in each school?

src/web/src/pages/CourseExplorer.vue
```diff
@@ -55,12 +103,37 @@ export default {
...
+      showCourseInfoModal: false,
     };
   },
   async created() {
     getCourses(this.selectedSemester).then((courses) => {
       for (const c of courses) {
+        if (!this.schoolsMajorDict[c.school]) {
+          this.schoolsMajorDict[c.school] = [c.department];
+        } else {
+          if (!this.schoolsMajorDict[c.school].includes(c.department)) {
```
123joshuawu, line 133: could use javascript Set instead of list, then you can just do `this.schoolsMajorDict[c.school].add(c.department)`

src/web/src/utils.js
```diff
@@ -163,6 +163,9 @@ export const withinDuration = (s1, s2) => {
 
 export const generateRequirementsText = (prereqs, coreqs, raw) => {
   let text = [];
+  if (!prereqs && !coreqs) {
+    return "This course has no prerequisites";
```
123joshuawu, line 167: I believe you also need to check that raw is an empty string as well. there are cases where we are unable to parse the prereqs and coreqs

123joshuawu:
Agree with @jshom 

That will be the object of the next pr

src/web/src/pages/CourseExplorer.vue
IkeLyons: That doesn't work just because the schools are pulled out in order, so if you sort by number of courses the first column would always have the two biggest schools. If we did it like that then HASS and Science would be in the first column which would be way bigger than the second.

123joshuawu: Makes sense carry on

PR #197, opened 2020-08-13
Darkmode

### LINKED ISSUE #196: Feature Request — Dark Mode
**Job**
Add a toggle-able dark mode. 

Use CSS to modify each component's style and use vue-cookies to store the user's style preference.

**Context**
N/A

aidan-lane:
**Issue**
> closes #196 

> None

**Test Modifications**
> None

**Test Procedure**
> None

**Photos**
Screenshot_2020-08-13 YACS

Show before and after, capture screenshot/gif of finished feature/bug

**Additional Info**
Make sure to edit dark.scss whenever a new component is added. Not doing so would cause things to look out of place. It also might be worth looking into different ways of organizing the css, as it is quite long.


src/web/src/main.js
```diff
@@ -13,10 +14,24 @@ Vue.config.productionTip = false;
 
 Vue.use(VueRouter);
 Vue.use(VueCookies);
+Vue.use(Vuex);
+
+const store = new Vuex.Store({
```
123joshuawu, line 19: I'd suggest moving the `store` declaration to its own file, then import store here.
Can put it in `src/web/src/store/index.js` then you do `import store from "../store"` or something along those lines

src/web/src/main.js
```diff
@@ -13,10 +14,24 @@ Vue.config.productionTip = false;
...
+  state: {
+    darkMode: false,
+  },
+  mutations: {
+    setDarkMode(state, bool) {
```
123joshuawu, line 24: I'd also suggest using constants for mutation and action types. Here's the official docs on it: https://vuex.vuejs.org/guide/mutations.html#using-constants-for-mutation-types. You don't need to put the constants in a separate file, having them at the top of the `index.js` will work fine

src/web/src/main.js
```diff
@@ -13,10 +14,24 @@ Vue.config.productionTip = false;
...
+    },
+  },
+});
 
 Vue.$cookies.config("7d");
 
 new Vue({
   render: (h) => h(App),
   router,
+  store: store,
```
123joshuawu, line 36: js tip: if the key and variable that contains the value have the same name, you can just put the variable name and js will infer the key name. So e.g. right above on line 35, that is equivalent to `router: router`

src/web/src/components/Header.vue
```diff
@@ -105,16 +114,17 @@ export default {
...
         email: "",
         password: "",
       },
-
       isLoggedIn: false,
       sessionID: "",
       userName: "",
       showForm: true,
       semesterOptions: [],
+      darkModeEnabled: this.$cookies.get("darkMode"),
```
123joshuawu, line 127: Probably want to use `$store.state.darkMode` instead of cookies here

src/web/src/components/Header.vue
```diff
@@ -85,6 +85,13 @@
           </b-modal>
         </div>
       </b-navbar-nav>
+      <b-form-checkbox
+        class="dark-mode-switch"
+        v-model="darkModeEnabled"
```
123joshuawu, line 90: other option instead of using `v-model`, you can use `:value=<initial value>` and `@input=<callback when value changes>`. here's the official docs: https://vuex.vuejs.org/guide/mutations.html#using-constants-for-mutation-types. using `v-model` works but adds an extra step in the process

src/web/src/App.vue
```diff
@@ -8,10 +8,16 @@
 export default {
   name: "App",
   components: {},
+  created() {
+    if (this.$cookies.get("darkMode") == "true") {
+      this.$store.commit("TOGGLE_DARK_MODE");
```
123joshuawu, line 13: now that you've declared the mutation type as a constant, you should import it and use it in lieu of the string

src/web/src/components/Header.vue
```diff
@@ -85,6 +85,14 @@
           </b-modal>
         </div>
       </b-navbar-nav>
+      <b-form-checkbox
+        class="dark-mode-switch"
+        :checked="$store.state.darkMode"
+        @change="$store.commit('TOGGLE_DARK_MODE')"
```
123joshuawu, line 91: here as well

PR #205, opened 2020-09-19
A simple fix for web container not being able to talk to api container

RichtXO:
**Issue**

On the Windows dev environment, the web container was getting a proxy error, wasn't able to talk to the API container. 
It could not `proxy request /api/bulkCourseUpload from localhost:8080 to http://localhost:5000. (ECONNREFUSED).`

**Fix**
1. Changed the location of the API target from `http://localhost:5000` to `http://yacs_api:5000` since the NGINX set proxy_pass to that.
2. Added a `secure` field to not proxy everything to the API target.

**Additional Info**

https://medium.com/@bryantjiminson/solving-proxy-error-could-not-proxy-request-xxx-from-yyy-from-local-reactjs-app-to-nodejs-app-f28f3548afb9

### DIFF
--- a/src/web/vue.config.js
+++ b/src/web/vue.config.js
@@ -6,7 +6,7 @@ module.exports = {
     disableHostCheck: true,
     proxy: {
       "/api": {
-        target: process.env.YACS_API_HOSTNAME || "http://localhost:5000",
+        target: process.env.YACS_API_HOST || "localhost:5000",
         changeOrigin: true,
         secure: false,
       },
--- a/src/web/vue.config.js
+++ b/src/web/vue.config.js
@@ -6,8 +6,9 @@ module.exports = {
     disableHostCheck: true,
     proxy: {
       "/api": {
-        target: "http://localhost:5000",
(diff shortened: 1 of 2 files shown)


src/web/vue.config.js
```diff
@@ -6,8 +6,9 @@ module.exports = {
     disableHostCheck: true,
     proxy: {
       "/api": {
-        target: "http://localhost:5000",
+        target: "http://yacs_api:5000",
```
123joshuawu, line 9: The API url will be `yacs_api` only when running the whole application in docker. When running outside of docker, this needs to be `localhost`. 

To do this, we can allow overriding the hostname of the api through an environment variable like `YACS_API_HOSTNAME` but provide a default of `"localhost"` if `YACS_API_HOSTNAME` is not specified. Then, in `docker-compose.development.yml`, set the environment variable for the `yacs_web` service.

123joshuawu (changes requested):

123joshuawu (approved):

PR #213, opened 2020-10-03
Fix Schedule Conflict Message Feature Solution:

### LINKED ISSUE #211: Scheduling - Schedule Conflict Message
**Job**
When I get a schedule conflict, I want to be informed of which courses have conflicts and at what time block so that I can easily find it in the schedule.

_Current Message:_
Screen Shot 2020-10-02 at 3 30 37 PM

_Change to Message:_
**Header:** Cannot add CONTEMPORARY DESIGN APPROACHES
**Message:** CONTEMPORARY DESIGN APPROACHES [CRN] conflicts with OTHER COURSE [CRN] {color icon} at 10:00AM
_keep color icon the same (the blue box in the screenshot example)_

venkatrsrinivas:
Hello Everyone!
This Is My Pull Request Concerning The Updated Schedule Conflict Message Feature. 
Below You Will Find All Important Details: 

**Formal Issue**

This Pull Request Should Ideally Closes #211: Scheduling - Schedule Conflict Message. 

**Full Testing Procedure**

*Sample Example(s) To Test Functionality:*
> 1. Run Development Environment w/ Loaded Fall 2020 Semester Data.
> 2. Add 'Machine Learning From Data'.
> 3. Now, Try Adding 'Networking In The Linux Kernel'. 
> 4. Should Be Notified w/ Updated Conflict Error Message Data. 

**All Relevant Photos**

Original:

94968893-58d09600-04c7-11eb-8d2c-da89e0d9787d

Final: 

Screen Shot 2020-10-02 at 22 10 37

**Additional Information:**

NOTE: The Hardest Challenge In This Issue Was Determining The Existing Course Title That Conflicted w/ The Input Course Trying To Be Added, Since We Did Not Pass Course Data, But Originally Just Course Session Data. 

Thank You!


venkatrsrinivas:
Just Added Those Final Changes That Jacob Mentioned From Issue, Concerning The Time Format, ...

venkatrsrinivas:
Here Is The Latest Version For Reference, But Please Feel Free To Test On Your Own. 
Screen Shot 2020-10-02 at 23 05 58

src/web/src/pages/CourseScheduler.vue
```diff
@@ -274,10 +274,25 @@ export default {
...
+                  currentSection
+                ) {
+                  return currentSection.crn == err.existingSession.crn;
+                });
+                //Found Course w/ Appropriate CRN.
+                if (isFound) {
+                  return currentCourse;
+                }
+              });
+              //Call Helper Function To Actually Notify Client.
```
jshom, line 290: No need to describe what it does here, name has enough information here.

src/web/src/pages/CourseScheduler.vue
```diff
@@ -340,10 +355,25 @@ export default {
...
+              currentSection
+            ) {
+              return currentSection.crn == err.existingSession.crn;
+            });
+            //Found Course w/ Appropriate CRN.
+            if (isFound) {
+              return currentCourse;
+            }
+          });
+          //Call Helper Function To Actually Notify Client.
```
jshom, line 371: This seems like a replica from the lines 277-290, lets make a method for this to remove a bit of duplication.

src/web/src/mixins/NotificationsMixin.js
```diff
@@ -10,27 +11,41 @@ import { getBackgroundColor, getBorderColor } from "@/services/ColorService";
...
+     * @param {CourseSession} existSession = Course Session Data For Existing Course.
      */
-    notifyScheduleConflict(course, conflictSession, subsemester) {
+    notifyScheduleConflict(addCourse, existCourse, addSession, existSession) {
+      //NOTE: If Sub-Semester Data Desired, Need To Add As Parameter +
+      //Pass From CourseScheduler.vue.
+      // const vNodesMsg = this.$createElement("p", { class: ["mb-0"] }, [
+      //   `${subsemester.display_string}: Conflict With ${existSession.crn}
+      //         - ${existSession.section} `,
+      //Format Time Into Human-Readable Via Moment (e.g., 10:00 AM):
```
jshom, line 27: For the note, this can be extracted out to the PR for reference.

jshom, line 27: No need to explain what moment.format does, this is implied in the method name.

Generally I'd prefer to stay away from these single line comments that explain what the next line does, usually adds more lines of comments without adding much value as long as the method names are descriptive enough.

What do you think?

jshom (approved):
Besides a bit of duplication and cleaning up some comments, looks good to me!

src/web/src/pages/CourseScheduler.vue
123joshuawu: Agree, this can be extracted to `util.js`

src/web/src/pages/CourseScheduler.vue
```diff
@@ -274,10 +274,25 @@ export default {
...
+              //Compute Existing Course That Caused Conflict.
+              const existCourse = this.courses.find(function (currentCourse) {
+                //Search For Desired CRN In Current Course.
+                var isFound = currentCourse.sections.find(function (
+                  currentSection
+                ) {
+                  return currentSection.crn == err.existingSession.crn;
+                });
+                //Found Course w/ Appropriate CRN.
+                if (isFound) {
```
123joshuawu, line 286: an alternative/more semantic way of doing this is to use js array method `some` for determining if the current section has the `err.existingSession`

123joshuawu (commented):
Agree with jshom's comments, left some js tips

venkatrsrinivas:
Hello Jacob + Josh,
Thank you for the comments!
I just made those changes, and kindly check them out and let me know if anything else is necessary.
Also, as far as I can, the functionality should still be the same, but if you would like to verify as well, that would be great!

src/web/src/utils.js
```diff
@@ -210,6 +210,22 @@ export const addDays = (date, days) => {
   return result;
 };
 
+/**
+ * Export Course Data Given CRN Value.
+ */
+export const conflictCourse = (allCourseData, inputCRNValue) => {
```
123joshuawu, line 216: nit: I'd suggest calling this `findCourseByCourseSessionCrn`

src/web/src/utils.js
```diff
@@ -210,6 +210,22 @@ export const addDays = (date, days) => {
...
+  return allCourseData.find((currentCourse) => {
+    //Search For Desired CRN In Current Course.
+    let isFound = currentCourse.sections.some((currentSection) => {
+      return currentSection.crn === inputCRNValue;
+    });
+    //Found Course w/ Appropriate CRN.
+    if (isFound) {
```
123joshuawu, line 223: js tip: you could just return `isFound` instead of checking it and returning `currentCourse`, so something like this `allCourseData.find((currentCourse) => currentCourse.sections.some((currentSection) => currentSection.crn === inputCRNValue))`

123joshuawu (approved):
A couple suggestions, otherwise approved, great work!

src/web/src/utils.js
venkatrsrinivas: Hello Josh. Thank you for this suggestion. However, I just tried this out and unfortunately it did not work. The reason I actually implemented it this way is due to the fact that .some() returns True/False and we are interested in the corresponding currentCourse that made it return True for .some().

src/web/src/utils.js
venkatrsrinivas: Sounds good to me. I completely agree since more descriptive function names would make everything more clear. It should be updated by now.

venkatrsrinivas:
Hello Everyone. I was wondering, from an RCOS Perspective, if the "Squash + Merge" would prevent everyone (i.e., the Graders) from seeing all of my commits on the Master Branch? I know Open-Source Contributions are important and so if all of my individual commits are squashed into one, would this not be bad?

123joshuawu:
You can put a link to the pull request so the graders can see your commits.

And the `codecov` checks are failing because we added code without adding tests, so our test coverage decreased by `0.24%` which is fine

PR #234, opened 2020-10-13
Filtered out courses with no sections (in course list)

### LINKED ISSUE #220: UX Bug — Handle courses with no sections
**Describe the bug**

Some courses that do not have sections like ADMN-1030 show up in RPI Spring 2020 but upon clicking them, nothing happens. We need to inform students on what is/has happening/happened

Screen Shot 2020-10-09 at 1 03 44 PM

**To Reproduce**
1. Load Spring 2020 RPI
2. Select ADMN-1030, _notice the lack of feedback_

**Expected behavior**

This is a bit up for debate here, we have a few options:
1. Hide courses with no sections, this would eliminate this problem and cure future headaches
2. Provide a message like we do with course conflicts
3. Provide messaging in the course listing itself

Maybe we do something like this for no sections? (i.e  Provide messaging in the course listing itself)
Screen Shot 2020-10-09 at 1 12 05 PM

Note: before working on this, make sure this description is updated with the precise work definition

nick-lennox:
**Issue**

> closes #220 

Filtered out courses from the loaded CSV that do not contain any sections.

n/a

**Test Procedure**

 1. Load Spring 2020 CSV
 2. Notice courses on course list
 3. Ensure all courses in list have sections (show on schedule w/ a time)

### DIFF
--- a/docs/dev-on-mac-and-linux.md
+++ b/docs/dev-on-mac-and-linux.md
@@ -25,3 +25,12 @@ See install guides here if you are missing any:
 From the project root, run `./scripts/dev-start.sh`
 
 _See dev script here: https://github.com/YACS-RCOS/yacs.n/blob/master/scripts/dev-start.sh_
+
+## Dev Clear Volumes
+
+If any of the following occur
+
+- database schema is changed
+- web dependencies are changed
+
+From the project root, run `./scripts/dev-clear-volumes.sh`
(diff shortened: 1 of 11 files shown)


jshom:
Nice!

One thing, noticing that this would filter sectionless courses in explore page as well.
In fall-2020 for instance, seeing course ADMIN-1030, this course should still show up in explore.

Can you parametrize this call so it only filters from CourseScheduler and not Explore.

Thanks!

jshom (approved):
lgtm!

PR #251, opened 2020-10-27
remove cypress and coverage from ci

123joshuawu:


jshom:
Can you remove the lines instead of comment, we can recover the change with a git revert

jshom (approved):

PR #313, opened 2020-11-11
Fix spring update workflow

123joshuawu:
If we want to have more detailed diff for the csv (like line by line changes), just lmk and I'll add it


jshom (approved):
lgtm!

marchdan (approved):

Issue #302, opened 2020-11-09
Feature Request — Prerequisites should say whether they're an OR or an AND


benjaminrsherman:
**Job**
Right now, when a course has multiple prerequisites they are just displayed with a comma between them.  For some classes, you must have all of the prerequisites (e.g. Xinformatics requires CSCI 1200 and CSCI 4350).  For others, you can have any of the prerequisites (e.g. Computability and Logic requires PHIL 2140 or CSCI 2200), and for some there is a combination (e.g. Operating Systems requires CSCI 2300 and either CSCI 2500 or ECSE 2660).  The current approach makes it impossible to tell which combinations of the prerequisites are actually required, rendering the list of them almost useless.

**Describe a possible solution**
Instead of saying "Requires completion of X, Y, Z", YACS should say something like "Requires completion of X and Y or Z".

IkeLyons:
Closed by pull request #323

PR #323, opened 2020-11-13
Prereq links

### LINKED ISSUE #283: Feature Request — Add links to course page of prerequisite/co-requisite courses
**Job**
If a course has pre/co-requisites, I should be able to navigate those course pages. 

CSCI-1200 should be a link that brings me to the CoursePage for CSCI-1200
Screen Shot 2020-10-31 at 1 29 29 AM

**Context**

IkeLyons:
**Issue**

Closes #283, #302 

N/A

**Test Modifications**
N/A

**Test Procedure**
1. Open any course page with prerequistes
2. Notice the links and distinctions between "and"s and "or"s

**Photos**
Before:
Screen Shot 2020-11-13 at 3 49 18 PM

After:
Screen Shot 2020-11-13 at 3 48 11 PM

**Additional Info**
Also ended up deleting some more legacy code


jshom:
Screen Shot 2020-11-13 at 4 24 09 PM

oh. ye

src/web/src/pages/CoursePage.vue
```diff
@@ -70,10 +62,25 @@ export default {
...
   },
   computed: {
-    prereqTree() {
-      let treeData = {};
-      treeData.name = this.courseName;
-      return treeData;
+    transformed() {
+      let precoreqtext = this.courseObj.raw_precoreqs;
+      if (precoreqtext === null) {
+        return "This course has no prerequisites";
```
jshom, line 68: Maybe we change this line over to "No information on pre/corequisites" wyt?

src/web/src/pages/CoursePage.vue
```diff
@@ -70,10 +62,25 @@ export default {
...
+      }
+      const regex = /([A-Z]){4}( )([0-9]){4}/g;
+      while (precoreqtext.search(regex) != -1) {
+        let index = precoreqtext.search(regex);
+        let beforetext = precoreqtext.slice(0, index);
+        let dept = precoreqtext.slice(index, index + 4);
+        let cn = precoreqtext
```
jshom, line 75: lets rename `cn` to `course_number`, had to think for a sec to figure this one out

jshom (approved):
sweet! and maybe rename and message change as well

src/web/src/pages/CoursePage.vue
```diff
@@ -70,10 +62,25 @@ export default {
...
+          .slice(index, index + 9)
+          .split(" ")
+          .join("-");
+        let link = '<a href="/explore/'.concat(dept, "/", cn, '">', cn, "</a>");
```
123joshuawu, line 79: Prefer if we can use `router-link`

123joshuawu (changes requested):

src/web/src/pages/CoursePage.vue
IkeLyons: I was trying to use router link but it didn't seem to like it with the vthml in the template. I think this is because the vbind:to directive doesn't work inside it. After I tried using dynamic components but that was a whole other mess. There might be a way to get it to work but I couldn't figure it out and the a tags worked?

src/web/src/pages/CoursePage.vue
IkeLyons: a little non-committal ;) No thats much better

jshom:
_fixed a smol typo there_

src/web/src/pages/CoursePage.vue
123joshuawu: Yeah looks like vhtml doesnt work well with components. The only issue with `a` links is they will reload the page so if we start storing state in vuex this is gonna reset it. I guess there's no good solution here so we'll just keep it as it is

123joshuawu (approved):

Issue #294, opened 2020-11-06
Explore — Use breadcrumb component to standardize back functionality


jshom:
**Job**
When I am on a course page or subject page I want to know relatively where I am on the YACS page.

**Task**
Use VueBootstrap component: https://bootstrap-vue.org/docs/components/breadcrumb#breadcrumb

At the top of Department and Course pages.
i.e when on CSCI-4440, `Explore > CSCI > CSCI-4440`

**Example**

![Screen Shot 2020-11-15 at 5 39 19 PM](https://user-images.githubusercontent.com/11034221/99198910-9b91b900-2769-11eb-960b-54f2939d3092.jpg)

PR #328, opened 2020-11-16
294 - breadcrumb nav on explore pages

nick-lennox:
**Issue**

issue #294 
"When I am on a course page or subject page I want to know relatively where I am on the YACS page."

*Example:*
Added breadcrumb navigation on all subpages of /explore

**Test Procedure**

Go to YACS home -> explore, notice the row below the header that tells you where you are relative to the rest of the YACS pages

**Photos**

Screen Shot 2020-11-15 at 9 02 01 PM


123joshuawu (changes requested):
Sweet, looks really nice.

Just two things:
1) I think we don't need `Home` in the breadcrumb because `Explore` is the root page for all the explore pages
2) Can we use `to` instead of `href` for the links? This will avoid reloading the page when using the breadcrumbs

nick-lennox:
> 1. I think we don't need `Home` in the breadcrumb because `Explore` is the root page for all the explore pages

I feel like it may not be a bad idea to leave it in to make the navigation back to the main page a bit easier (we also don't necessarily need to call it 'Home'), but if you feel strongly about it I can definitely change it. For sure I'll change the hrefs though.

Thanks for the feedback!

123joshuawu (approved):
Sounds good to me

marchdan:
@nick-lennox @123joshuawu What about if we used "YACS" instead of "Home"? I do agree that Explore is kinda a root page, but explore is still a branch off of the root homepage. this is just a thought, I think either way works fine

nick-lennox:
> @nick-lennox @123joshuawu What about if we used "YACS" instead of "Home"? I do agree that Explore is kinda a root page, but explore is still a branch off of the root homepage. this is just a thought, I think either way works fine

I was thinking this would be a good solution too.

marchdan:
@nick-lennox Please resolve merge conflicts so this can be reviewed

src/web/src/pages/SubjectExplorer.vue
```diff
@@ -1,78 +1,78 @@
...
-            {{ course.title }}
-            <br />
-            {{ course.department }}
-            {{ course.level }}
-            <br />
-          </b-button>
-        </b-col>
-      </b-row>
-    </b-col>
+      <!-- left column of courses -->
```
marchdan, line 12: Is there a specific reason for the extra spacing here?

marchdan:
Is there a specific reason for the extra spacing change in this file?

https://github.com/YACS-RCOS/yacs.n/pull/328/files#diff-a09f667c120da1ee7f101c7058309ff0ed6bbfd9432a1077ad7c0cd480d00257

marchdan (approved):
Looks good. Course explorer seems to have some weird loading issues, but I think there is a separate PR to fix that

marchdan:
@123joshuawu Could you double-check this? Looks good to me, but there have been changes since you approved

123joshuawu (approved):

PR #337, opened 2020-11-18
Performance - Handle duplicate /class request using Vuex course list

### LINKED ISSUE #325: Bug — Refetching /class
**Describe the bug**
We are requesting a single route for every course page load in YACS, this makes every page take atleast 0.5 second or so to load. This should only be fetched once on page load.

**To Reproduce**
1. Watch network tab in Firefox/Chrome Dev Tools
2. Go to explore
3. Go to CSCI
4. Go to any course

You'll see `/class` 4 times, its the same data.

**Expected behavior**
On page load, save the data, do not refetch until next page load

With this fix, the YACS should feel like 80% faster when going through the explore features

Screen Shot 2020-11-13 at 6 40 39 PM

IkeLyons:
**Issue**
closes #325 

N/A

**Test Modifications**
N/A

**Test Procedure**
1. open yacs
2. Click course explorer
3. Notice way faster load time!
4. Click any major
5. Notice way faster load time!
6. Click any course
7. Notice way faster load time!

**Photos**
N/A

**Additional Info**
Changes mainly consisted of adding the vuex state and then moving instance variables of pages to computed functions


123joshuawu:
@IkeLyons i think course lists needs to account for semesters as well?

IkeLyons:
@123joshuawu Just fixed that for the scheduler page. Currently explore and its subpages don't handle semester changes, so ill make that into its own issue

marchdan:
@IkeLyons This works well as far as I can see. The only slow-down in the explorer seems to occur when following pre/corequisite links. But it's very small.

marchdan:
I think #352 would be relevant to this PR

marchdan:
@IkeLyons If you think #352 should be separate from this, lmk and I can review this PR as is. Otherwise, I think #352 would be an easy additional fix for this PR

IkeLyons:
@marchdan yea I just figured the scope of that was kinda separate, I could pick up #352 but this pull deals more with the courselist itself and that aspect of the links is different.

marchdan:
> @marchdan yea I just figured the scope of that was kinda separate, I might pick up #352 but this pull deals more with the courselist itself and that aspect of the links is different.

No worries. I can review this PR later tonight and by the time it's merged maybe you'll have had some time to research how to handle the new issue.

marchdan:
@IkeLyons I'm getting some weird errors that may or may not be related to this PR. I'm going to try re-creating the temp deploy tomorrow and try again. The errors are listed below in case you have any idea what is causing them.

1. The default page is semester=null even after CSV has been uploaded and default semester is set to that semester. After one trip back to the root page it unsets the default semester back to nothing. Eventually, it unloads the uploaded CSV and says there are no courses again.

2. When I try and follow a pre-requisite or co-requisite link (a valid one) it hangs forever and then I can't navigate to any pages, even via changing the URL, until i return to root and go back to explorer and try again. 

These could be temp deploy system errors, but I can't test this PR while they are occurring.

IkeLyons:
Hm, yea that is strange. 
1. This doesn't seem to happen locally on my machine but I'm having the same thing happen on the temp deploy. Not sure what is causing it honestly. The way I can seem to get around it is by manually selecting the semester in the footer, but not sure why it doesn't like setting it as the default. I don't think I made any changes to the semester selection logic on the course scheduler, as it is just passed in as a prop from the student page where I haven't made changes.

2. I can't seem to recreate this one.

Stranger now, after writing all of the above I tried running the temp deploy again and now problem 1 isn't occurring but problem 2 sometimes happens. Seems like these are just temp deploy issues, but very strange nonetheless

marchdan:
@IkeLyons And you're trying on the temp deploy from today?

@jshom Any idea why these things might be happening in the temp deploy?

jshom:
Could be a caching issue, will investigate later this week.

123joshuawu:
Default semester is another thing that should be utilizing vuex, that might solve this issue as well

marchdan:
@IkeLyons Pining to remind you to settle file conflicts

IkeLyons:
@marchdan On this most recent temp deploy I am getting neither issue. Not sure what was up before but seems to be fixed now ¯\_(ツ)_/¯

123joshuawu:
We good to go on this pr? @marchdan @IkeLyons

marchdan (approved):
@IkeLyons I am still seeing some issues, but it doesn't seem to be fatal anymore. I think the course explorer loading issues on non-existent courses is attached to another issue. I can look into it more, but it shouldn't have an effect on this PR being approved.

src/web/src/pages/CoursePage.vue
```diff
@@ -128,6 +104,15 @@ export default {
       }
       return precoreqtext;
     },
+    courseObj() {
+      let course = {};
+      for (const c of this.$store.state.courseList) {
+        if (c.name === this.courseName) {
+          course = c;
```
123joshuawu, line 111: Probably good to break out of loop here, or can use `courses.find((course) => course.name === name)`

123joshuawu (approved):

src/web/src/pages/CoursePage.vue
IkeLyons: idk why I never think to use some of these built-in functions, thanks!

PR #345, opened 2020-11-20
Update Spring 2021 csv

123joshuawu:
rpi_data/spring-2021.csv | 845 ++++++++++++++++++++++++-----------------------
 1 file changed, 423 insertions(+), 422 deletions(-)


marchdan (approved):
There were some new classes added and some class sections dropped today, but all the one's I noticed matched up with the current class hours schedule.

PR #348, opened 2020-11-20
User session fixes for saved schedules

123joshuawu:
**Issue**

Discussed during meeting

N/A

**Test Modifications**

N/A

**Test Procedure**

Wink wink

**Photos**

N/A

**Additional Info**

N/A


marchdan:
😉 😉

123joshuawu:
@marchdan actually I think I will change it so we don't store cookie with user ID in the front end, can just use the session user id

marchdan:
> @marchdan actually I think I will change it so we don't store cookie with user ID in the front end, can just use the session user id

How will this affect the reloading of courses between log out and logging back in?

123joshuawu:
I don't think anything will change, it's just we won't have the user ID cookie, just the session cookie

marchdan:
> I don't think anything will change, it's just we won't have the user ID cookie, just the session cookie

When a user logs in, loadStudent courses uses the userID cookie to store the id before using it to retrieve the courses:
https://github.com/YACS-RCOS/yacs.n/blob/1e6ecaf36f8cedd7a62d11389489e95a6a1f4c34/src/web/src/pages/CourseScheduler.vue#L208-L242

123joshuawu:
Ye so now, it will call `getStudentInfo` and then if it returns courses, then load those, else, try load courses from savedCoursesCookie

marchdan:
> Ye so now, it will call `getStudentInfo` and then if it returns courses, then load those, else, try load courses from savedCoursesCookie

@123joshuawu That is what already happens... how will it get the userId to do that with?

marchdan:
> Ye so now, it will call `getStudentInfo` and then if it returns courses, then load those, else, try load courses from savedCoursesCookie

@123joshuawu So getStudentInfo returns the userID of the currently logged in user?

123joshuawu:
@marchdan hmm okay, i think I will need to change how user login/logout is handled on the frontend as well

jshom (approved):
sweeet

jshom:
btw, wyt about renaming the delete and post routes of /api/course to `/api/user/:id/course/:course` when I saw DELETE /api/course, it kinda scared me lol

123joshuawu:
@jshom yee, I'll change it

123joshuawu:
We're probably going to want to look at transitioning to https://flask-session.readthedocs.io/en/latest/ for handling sessions.

marchdan:
@123joshuawu What is the current status of this?

123joshuawu:
@marchdan should be good to go

marchdan:
> @marchdan should be good to go

@123joshuawu Can you resolve the conflicts so it can be reviewed?

PR #371, opened 2020-12-07
DB Migrations with Alembic

### LINKED ISSUE #315: DX — Use database migration tool for schema versioning and updates
**Job**
We need a way to update the database without dropping everything and rebuilding

This is particularly prevalent for the production deployment

Some sort of database migration tool that can generate database migrations for updating tables in place

Ideally this should:
1) On application start, check if sql table definitions match tables in database
2) If there is a difference, run migration script to update table(s) in database

Possible tools:
- Alembic https://alembic.sqlalchemy.org/en/latest/

**Context**

123joshuawu:
**Issue**

Closes #315 

Use SQLAlchemy ORM declarative class-based schema definitions instead of previous SQL table definitions. This sets up the possibility of ORM usage in the near future.

**Test Modifications**

Changed `TestData.clear_db` to use SQLAlchemy's `Metadata` to truncate all tables

**Test Procedure**

N/A

**Photos**

N/A

**Additional Info**

The introduction of db migrations is a major change, introducing a new process for updating the schema. 

One of the main changes was converting all of our `.sql` table definitions into SQLAlchemy ORM declarative schema classes in `src/api/tables`. I've tried to get them as equal as possible, but made a few additions when needed (e.g. tables not having primary keys).

Previously, schema updates were done by modifying the SQL files that defined the tables. These SQL files were loaded into the database and executed when the database started up. However, these initialization files were ignored if Docker Postgres detected an existing database (persisted through volumes). Thus, changing the schemas required deleting the docker volume. Not only was this an a potential tripping point for developers, it was unfeasible for production. 

Database migrations solves this issue. Alembic allows us to use migration scripts to update the database without any downtime or rebuilding or deleting data. Going an extra step by using SQLAlchemy to define our tables allows us to autogenerate migration scripts. 

Here is an outline of the new workflow for introducing a schema change.

1. Developer A changes the schemas in `src/api/tables`
2. Developer A runs `scripts/dev-create-migration.sh "new feature"` which automagically generates a database migration script `2020-12-10_new_feature.py` in `src/api/migrations/versions`
3. Developer A opens a PR
4. PR gets merged
5. Developer B pulls master
6. Developer B starts up the application with `scripts/dev-start.sh`
  When `yacs_api` is started, it will run `alembic upgrade head` which will check the database version, determine that one migration needs to be run, and then run `2020-12-10_new_feature.py`.

Developer B's database now contains the same schemas as Developer A's database. Notice how Developer B did not need to take any extra steps. 

We can also examine how a schema change would affect a production deployment. Continuing with the example above,

1. Steps 1-4 remain the same
5. Developer C SSHs into production deployment
6. Developer C pulls master
7. Developer C builds new `yacs_web` and `yacs_api` containers `docker-compose build`
8. Developer C stops current `yacs_web` and `yacs_api` containers
9. Developer C removes current `yacs_web` and `yacs_api` containers
10. Developer C starts new `yacs_web` and `yacs_api` containers
  Same as before, `yacs_api` will run the newest migration script on startup and update the database tables.

Notice how we never touched the database at all. This is because all of our schemas are now defined by `yacs_api`. 

When this feature is merged into master, developers will need to run `scripts/dev-build.sh`. This will let Alembic initialize the database with the version number. The production database will also need to be version stamped, so `alembic stamp head` will do the trick. 

That's the general overview of the process that I've envisioned. Some of this stuff should probably be documented somewhere within the repo. This could segue into our need for some guide/documentation collection of some sort, probably Github Wiki or something along those lines.


jshom (approved):
Yes :)

RichtXO (approved):

123joshuawu:
I will merge soon, will need to handle prod on merge

PR #388, opened 2020-12-24
production CD with self-hosted github runner

123joshuawu:
I've setup a self-hosted Github Action runner on the production VM so we can finally do CD through Github Actions.

The `prod-cd` workflow runs on master push or manual trigger. It uses the `yacs-prod` environment which contains the secrets for prod as well as configuration for manual verification of deployment. I've currently configured it such that deployments require manual approval (by me for now).

The `prod-cd` workflow will pull the new code from master, rebuild and restart `yacs_api` and `yacs_web` in the correct order. 

I've done some quick tests, you can go back in the commits or actions history to see the workflow in action. 

TLDR; CD process = push to master -> create new deployment -> manual approve -> redeploy production

### DIFF
--- a/.github/workflows/prod-cd.yaml
+++ b/.github/workflows/prod-cd.yaml
@@ -11,9 +11,36 @@ jobs:
     runs-on: [self-hosted, rpi, prod]
 
     steps:
+      # - uses: fountainhead/action-wait-for-check@v1.0.0
+      #   id: wait-for-ci
+      #   with:
+      #     token: ${{ secrets.GITHUB_TOKEN }}
+      #     checkName: CI
+
+      # - name: Check if ci was successful
+      #   if: steps.wait-for-ci.outputs.conclusion != 'success'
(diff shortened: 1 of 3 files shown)


.github/workflows/prod-cd.yaml
```diff
@@ -0,0 +1,63 @@
...
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+
+jobs:
+  deploy:
+    runs-on: [self-hosted, rpi, prod]
+    environment: yacs-prod
```
jshom, line 12: lets rename this to `yacs-rpi-prod` since rpi specific.
besides that, awesome!

jshom (approved):

.github/workflows/prod-cd.yaml
123joshuawu: Yep changed

PR #409, opened 2021-03-06
Credit Cap Feature Added

### LINKED ISSUE #382: Feature Request — Credit Cap
**Job**
Alert the student if they are past the credit cap. 

If the student wants to see how their schedule would look with different classes, that should be fine. However, if they're past the credit limit at RPI, there should be an alert, pop-up, or indication that they are past the credit limit. 

**Context**

yangytjason2000:
**Issue**
>close #382 
**credit cap**

Adding totalCredits() function to Schedule.js to help calculating credits sum. Adding a if-statement at the end of addCourseSection function to determine whether the credit exceeds credit cap of 21. Throw an error with type "Credit Conflict" when credit is over 21.
Also adding a function notifyCreditConflict() in NotificationMixin.js to be called in order to have a error message popup when error type "Credit Conflict" is throw and catch.


RichtXO:
Keep in mind that we also got grad students to look after too! I think their credit cap is 15. I guess we could assume user is UG when user isn't logged into the website. But you should check the account to see if it's UG or grad, and change the credit cap according.

src/web/src/controllers/Schedule.js
```diff
@@ -155,7 +155,15 @@ class Schedule {
       return false;
     }
   }
+  totalCredits() {
```
jshom, line 158: lets maybe rename this to `getMaxCreditsOfSelectedCourses`. wyt?

src/web/src/controllers/Schedule.js
```diff
@@ -155,7 +155,15 @@ class Schedule {
...
+    var array = Object.values(this.selectedCourses).map((c) => c.max_credits);
 
+    // Getting sum of numbers
+    var sum = array.reduce(function (a, b) {
+      return a + b;
+    }, 0);
+    return sum;
+  }
```
jshom, line 166: Wyt about simplifying this function a tad bit? This way we don't define abstract sum function and use it once
```js
return Object.values(this.selectedCourses)
  .map((course) => course.max_credits)
  .reduce((sum, max_credits) => sum += max_credits)

```

src/web/src/controllers/Schedule.js
```diff
@@ -245,6 +253,10 @@ class Schedule {
 
       this.selectedSections.push(courseSection);
       console.log(`Added new courseSection ${JSON.stringify(courseSection)}`);
+      if (this.totalCredits()+course.max_credits>21){
```
jshom, line 256: Perhaps we move the number `21` to admin panel. wyt @RichtXO @yangytjason2000?

src/web/src/controllers/Schedule.js
```diff
@@ -245,6 +253,10 @@ class Schedule {
...
+        console.error("Credit Cap");
+        throw {type:"Credit Conflict"};
```
jshom, line 258: Lets not throw an error here since this behavior is expected.

src/web/src/mixins/NotificationsMixin.js
```diff
@@ -48,5 +48,17 @@ export default {
...
+      //Format All Message Data For Conflict Message:
+      const vNodesMsg = this.$createElement("p", { class: ["mb-0"] }, [
+        'WARNING: Exceeding Credits Limit'
+        ]);
+      //Format All Header-Data For Conflict Message:
+      this.$bvToast.toast(vNodesMsg, {
+        title: `CREDIT CAP`,
+        variant: "danger",
+        noAutoHide: false,
+      });
```
jshom, line 61: Perhaps we link to admin panel configuration with RPI link? i.e. we configure in admin panel to link to http://catalog.rpi.edu/content.php?catoid=4&navoid=83.

So that In admin panel, we can set 1. credit warning limit 2. warning content

wyt?

src/web/src/controllers/Schedule.js
yangytjason2000: Then what do you suggest on how to handle this?

jshom: We can remove these lines

src/web/src/controllers/Schedule.js
yangytjason2000: I think this is a good way to implement the function. I will definitely modify the function.

src/web/src/controllers/Schedule.js
yangytjason2000: My problem is then how should I determine when to call the function I implemented in NotificationMixing.js? I call the function in NotificationMixing.js when I catch the error with type "credit conflict" when the function addCourseSection is called.

src/web/src/controllers/Schedule.js
RichtXO: I think we should utilize the type of students in the account when the user is signed in. I guess moving the number to the admin panel would be used for users who aren't logged in. I like that idea.

src/web/src/controllers/Schedule.js
jshom: you should be able to import NotificationMixin and directly call `NotificationMixin.notifyWarningCreditOverflow`

`notifyCreditConflict -> notifyWarningCreditOverflow`

123joshuawu: Unfortunately that won't work because the `NotificationMixin` needs to be applied to a `Vue` instance to function. This was my implementation mistake. Bubbling up an error to trigger the notification should work here, you'll just have to make sure that the course saving logic still runs.

PR #410, opened 2021-03-07
Automatically Switch to Dark Mode

### LINKED ISSUE #408: Feature Request — Automatically Switch to Dark Mode
**Job**
Automatically switch to dark mode for new users if their system color scheme is dark.

If there is no `darkMode` cookie, query the `prefers-color-scheme` to check the current color scheme, and add the cookie according to the query result.

**Context**
Dark mode may cause flicker since `.dark` class on `<body>` presents after the course list loading.

Viyerelu23333:
**Issue**
> closes #408

Untouched.

**Test Modifications**
Untouched.

**UI Modifications**
1. Removed the `switch` property of `b-form-checkbox`.
2. Moved out the moon icon from `#darkmode-toggle-form`.
3. Deleted the `<div>` near the moon icon.
4. Added `@dblclick` event to the moon icon.
5. Added two toast notifications.

**Test Procedure**
1. Double-click the moon icon to remove `darkMode` cookie (follow device's theme):
  The checkbox should in `indeterminate` state. 
2. Change the system theme to dark/light, the page will change color scheme *immediately*.
3. Click on the checkbox to specify dark/light mode.
  The checkbox should in `checked` state if the page is in dark mode, and `unchecked` for light mode.
  The color scheme should follow the checkbox even if the system color is changed.
4. Double-click the moon icon to follow the device's theme again, and remove cookie.

**Photos**
Indeterminate
Checked
Unchecked

**Additional Info**
Modified, added some functions in `store/index.js`, which may impact the current usage of cookies.
If it has significant impact, the previous commit `d50be64` is a brutal force implementation, which does not change any current functions.

**Current Status**
~Working on automatically change color scheme without refreshing.~
~This works best with pure CSS, which needs refactoring.~
Never mind, finished...


RichtXO (approved):
Just make the toasters appear on the top right! Other than that, I like it!

Viyerelu23333:
> 
> 
> Just make the toasters appear on the top right! Other than that, I like it!

Fixed.

src/web/src/App.vue
```diff
@@ -13,17 +13,30 @@ export default {
...
+
+    if (window.matchMedia &&
+        window.matchMedia(darkModeQueryBody).media !== 'not all') {
+      const queryColorScheme = window.matchMedia(darkModeQueryBody);
+      queryColorScheme.addEventListener('change', () => {
+        if (this.$cookies.get("darkMode") === null) {
+          this.$store.commit(TOGGLE_DARK_MODE);
+        }
+      });
+    }
```
jshom, line 31: From what I understand, it looks like the first `if` block is checking the cookies to see if the user has an existing color-scheme preference and the second `if` block is setting the colorscheme to the device preference. right?

If so, lets maybe refactor a bit. Here's a few ideas so that we can get to:
```js
async created() {
  this.syncDarkMode();
  ...
}
syncDarkMode() {
  if (this.isDarkModeCookieSet()) {
     this.syncDarkModeWithCookie();
  } else {
     this.syncDarkModeWithDevicePreferance();
  }
}
```
this way the code reads a bit simpler and we can seperate the concerns of checking if the dark mode is set, linking to cookie and matching device preference. 

wyt?

src/web/src/components/Header.vue
```diff
@@ -113,11 +110,36 @@ export default {
...
     toggle_style() {
+      if (this.$cookies.get("darkMode") === null) {
+        this.$bvToast.toast(
+          `Double click moon icon to follow device's color scheme.`, {
+            title: 'Color Scheme Changed',
+            autoHideDelay: 2000,
+            noHoverPause: true,
+            variant: "info",
+        });
+      }
```
jshom, line 126: wyt about extracting these toasts into named functions? could make `toggle_style()` and `toggle_default()` a little easier to read

src/web/src/App.vue
Viyerelu23333: Thanks for your comments!
Yes, I think your refactoring ideas are great.  Semantic function names are intuitive.  

There are still two points:
1. The first `if` checks the cookie or current scheme, and the second `if` checks the native support of color scheme: both `if` blocks will run, so it cannot be considered as a `if-else` block.
2. [Most browsers](https://caniuse.com/mdn-api_window_matchmedia) support `window.matchMedia`, so I deleted the testing for capabilities.

How about this:  `App.Vue`
```javascript
  async created() {
    this.darkModeInit();

    const querySemester = this.$route.query.semester;
    this.selectedSemester =
    // ......
  },
  methods: {
    darkModeInit() {
      this.syncColorScheme();
      this.registerColorSchemeListener();
    },
    syncColorScheme() {
      if (this.$cookies.get("darkMode")  === "true" ||
          (this.$cookies.get("darkMode") ===  null  &&
           window.matchMedia("(prefers-color-scheme: dark)").matches)) {
        this.$store.commit(TOGGLE_DARK_MODE);
      }
    },
    registerColorSchemeListener() {
      window.matchMedia("(prefers-color-scheme: dark)").addEventListener(
        'change', () => {
          if (this.$cookies.get("darkMode") === null) {
            this.$store.commit(TOGGLE_DARK_MODE);
        }
      });
    }
  },
```

Any ideas?

src/web/src/components/Header.vue
Viyerelu23333: Totally agree.  These `toast` calls will be wrapped as functions and moved to the bottom of `methods:` on the next commit.

src/web/src/App.vue
jshom: Reads much better! Lets take it one step further.

> **Describe a possible solution**
If there is no darkMode cookie, query the prefers-color-scheme to check the current color scheme, and add the cookie according to the query result.

Lets only read the preferance once with `.matches` and not keep the listener open. This change would sit more inline with the UX that we are looking for from the issue definition

Codewise, we'd no longer need the eventListener as we'd only take the value once that you can wrap in a function `getDarkModeDevicePreferance`.

Then, I think we can follow the earlier logic that reads at the same abstraction level as the issue statement
```js
sync created() {
  this.syncDarkMode();
  ...
}
syncDarkMode() {
  if (this.isDarkModeCookieSet()) {
     this.syncDarkModeWithCookie();
  } else {
     this.syncDarkModeWithDevicePreference(); // get pref, set dark mode, and save cookie
  }
}
```

Also small note on this predicates like this (not sure if itll be the same one):
```js
(this.$cookies.get("darkMode")  === "true" || (this.isFollowingDeviceColor() && mq.matches)
```
but better to give it a function so that when reading you can stay at one level of abstraction (figuring out where darkmode preferances are coming from) and not worry about mq.matches and what that means and what a JS MediaQuery object is.

Lastly, strings `"darkMode:` and the `"(prefers-color-scheme: dark)"` can be extracted out like
https://github.com/YACS-RCOS/yacs.n/blob/master/src/web/src/controllers/Schedule.js#L7-L8
so that we catch string mismatches as compile time errors. You can throw these strings up to the top of the class as something like `App.COOKIE_DARKMODE` and `App.MQ_COLORSCHEME` or `App.MEDIAQUERY_COLORSCHEME`

wyt? or should we keep the responsiveness to colorscheme changes?

src/web/src/components/Header.vue
jshom: btw, no need to worry about showering commits in for prs. they'll get squashed as we merge into master.

src/web/src/App.vue
123joshuawu: I'm way out of the loop here, but instead of only using `prefers-color-scheme: dark` when the `darkMode` cookie is null, could we just update the `darkMode` cookie state to match the `prefers-color-scheme: dark` state if it changes?

Viyerelu23333: > Lets only read the preferance once with .matches and not keep the listener open. This change would sit more inline with the UX that we are looking for from the issue definition
> 
> Codewise, we'd no longer need the eventListener as we'd only take the value once that you can wrap in a function getDarkModeDevicePreferance.
> 
> Then, I think we can follow the earlier logic that reads at the same abstraction level as the issue statement [...]

Sorry, I forgot to update the feature request.  My bad.  That was the original idea, and it only covers the change for new users who don't have cookies.  On actual coding I found it is easy to implement a responsive color scheme as a third state...  Personally I would like to preserve this feature.  

> Also small note on this predicates like this (not sure if itll be the same one):
> 
> ```javascript
> (this.$cookies.get("darkMode")  === "true" || (this.isFollowingDeviceColor() && mq.matches)
> ```
> 
> but better to give it a function so that when reading you can stay at one level of abstraction (figuring out where darkmode preferances are coming from) and not worry about mq.matches and what that means and what a JS MediaQuery object is.

Agree.  I thought this sentence shows a clear relationship of the conditions...  But not optimal for reading it.  
I will wrap it before the next push.

> Lastly, strings `"darkMode:` and the `"(prefers-color-scheme: dark)"` can be extracted out like [/src/web/src/controllers/Schedule.js: L7-L8](https://github.com/YACS-RCOS/yacs.n/blob/master/src/web/src/controllers/Schedule.js#L7-L8) so that we catch string mismatches as compile time errors. You can throw these strings up to the top of the class as something like `App.COOKIE_DARKMODE` and `App.MQ_COLORSCHEME` or `App.MEDIAQUERY_COLORSCHEME`

Do you mean `state.darkMode` in `store/index.js`?  Not sure if this will affect current usage of cookies... 
Also, not sure if `App.Vue` has a `class`...

Viyerelu23333: > I'm way out of the loop here, but instead of only using `prefers-color-scheme: dark` when the `darkMode` cookie is null, could we just update the `darkMode` cookie state to match the `prefers-color-scheme: dark` state if it changes?

It is a bit confusing.  
If I interpret your idea correct, `darkMode` cookie has a string for saving the current user's option for color scheme.  

Also there is a boolean `state.darkMode`, which will record the current site color scheme.  This will change on each load, if the cookie was set to dark mode; or the user did not specify the scheme, it will change to follow the current device preference.

jshom: Yep you're right, extracting out into `store/index.js` should do it.

Shouldn't change the behavior, only a code change

Proposed change goes from:
```js
return this.$cookies.get("darkMode") === null;
```
to
```js
return this.$cookies.get(App.COOKIE_DARKMODE) === null;
```
so easier to take advantage of autocomplete and reduce typos/silent errors

Viyerelu23333: > extracting out into `store/index.js` should do it where you'd import `COOKIE_DARKMODE`.
> Shouldn't change the behavior, only a code change
> 
> Proposed change goes from:
> ```javascript
> return this.$cookies.get("darkMode") === null;
> ```
> to
> ```javascript
> return this.$cookies.get(COOKIE_DARKMODE) === null;
> ```
> so easier to take advantage of autocomplete and reduce typos/silent errors

Perhaps something like this code below?
`store/index.js`: `export const COOKIE_DARK_MODE = "darkMode";`
Same `import` will apply to `Header.Vue` too.

<details>
  <summary><code>App.Vue</code></summary>

```javascript
import { TOGGLE_DARK_MODE, SET_COURSE_LIST, COOKIE_DARK_MODE } from "@/store";
export const DARK_MODE_MEDIA_QUERY = "(prefers-color-scheme: dark)"; 
export default {
  methods: {
    syncColorScheme() {
      if (this.isCookieOrDeviceInDarkMode()) {
        this.$store.commit(TOGGLE_DARK_MODE);
      }
    },
    registerColorSchemeListener() {
      window.matchMedia(DARK_MODE_MEDIA_QUERY).addEventListener(
        "change", () => {
          if (this.isFollowingDeviceColor()) {
            this.$store.commit(TOGGLE_DARK_MODE);
        }
      });
    },
    isFollowingDeviceColor() {
      return this.$cookies.get(COOKIE_DARK_MODE) === null;
    },
    isCookieOrDeviceInDarkMode() {
      return this.$cookies.get(COOKIE_DARK_MODE) === "true" ||
              (this.isFollowingDeviceColor() && 
               window.matchMedia(DARK_MODE_MEDIA_QUERY).matches);
    }
  },
}
```

</details>

RichtXO (approved):
Good work! LGTM

123joshuawu (approved):

Issue #418, opened 2021-03-09
Bug — Subsemester selection doesn't always show up


marchdan:
**Describe the bug**
Subsemester filter does not show up when Summer is navigated to from the footer if not default

**To Reproduce**
1. Go to yacs.cs.rpi.edu
2. Click on navigate to SUMMER 2021 using the site footer
3. Look for sub-sememester course filters

**Expected behavior**
Course filter and schedule filter should show up for sub-semesters

**Screenshots**
![Screenshot 2021-03-09 123927](https://user-images.githubusercontent.com/35609442/110514017-d7fc7280-80d4-11eb-9587-d0e62f639fea.png)

**Additional context**
I believe this to be a Vuex issue and may be fixed by pending PR.

RichtXO:
Update: This bug was caused by someone who uploaded the wrong version of the courses CSV :) (my bad)

123joshuawu:
Should be fixed by #425?

PR #424, opened 2021-03-12
Nginx cache Solution

RichtXO:
_This is a new PR to the old [PR](https://github.com/YACS-RCOS/yacs.n/pull/422). It just stopped looking at updated pushes that I made. IDK why it stopped, nor do I want to put the time to figure that out_

Issue
A possible fix to issue #418 

Sometimes, when updating semester courses, it would still display the old course because it's not actually pinging origin server; it's taking it from the nginx proxy server. Hopefully, shorting the cache age would help. Also took out gzip and implemented Brotli.

Additional Info

If Brotli is not desired, Gzip can be used.

### DIFF
--- a/src/web/nginx.conf
+++ b/src/web/nginx.conf
@@ -20,6 +20,7 @@ http {
   map $sent_http_content_type $expires {
       text/css                   30d;
       application/javascript     30d;
+      application/json            5m;
   }
 
   proxy_cache_path /tmp/micro_cache levels=1:2 keys_zone=micro_cache:100m max_size=100m inactive=300s;
@@ -68,12 +69,14 @@ http {
     if ($request_method !~ ^(DELETE|GET|POST)$ ){
       return 405;
     }
-    # prevent clickjack attacks
-    add_header X-Frame-Options "SAMEORIGIN";
-    # prevent XSS attacks
-    add_header X-XSS-Protection "1; mode=block";
-    # Configuring HSTS
-    add_header Strict-Transport-Security 'max-age=31536000; includeSubDomains; preload' always;
+    
+    # Headers
+    add_header X-Frame-Options "SAMEORIGIN";                                                      # prevent clickjack attacks
+    add_header X-XSS-Protection "1; mode=block";                                                  # prevent XSS attacks
+    add_header Strict-Transport-Security 'max-age=31536000; includeSubDomains; preload' always;   # Configuring HSTS
+    add_header CC-X-Request-ID $request_id;
+    add_header X-Cache_Status $upstream_cache_status;
+
     # Secure Diffie-Hellman for TLS
     ssl_dhparam /etc/nginx/cert/${HOST}.pem;
 
@@ -81,18 +84,13 @@ http {
     ssl_session_timeout 10m;
     
     # Enable session tickets
-    ssl_session_tickets off;
+    ssl_session_tickets on;
 
     # OCSP stapling
     ssl_stapling on;
     ssl_stapling_verify on;
     resolver 8.8.8.8 8.8.4.4 208.67.222.222 208.67.220.220 valid=60s;
     resolver_timeout 2s;
-
-    # More Headers
-    add_header CC-X-Request-ID $request_id;
-    add_header X-Cache_Status $upstream_cache_status;
-
 
     # simple secure admin panel, will change later
     location ~* ^/admin {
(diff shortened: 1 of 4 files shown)


src/web/nginx.conf
```diff
@@ -38,13 +39,23 @@ http {
...
-    gzip_comp_level 6;
-    gzip_buffers 16 8k; 
+    # gzip on;
+    # gzip_static on;    
+    # gzip_proxied  any;
+    # gzip_vary on;
+    # gzip_comp_level 6;
+    # gzip_buffers 16 8k;
+    # gzip_min_length 600;
+    # gzip_types image/jpeg image/bmp image/svg+xml text/plain text/css application/json application/javascript application/x-javascript text/xml application/xml application/xml+rss text/javascript image/x-icon;
```
jshom, line 49: Remove comments plz, other than that lgtm

jshom (approved):

src/web/nginx.conf
```diff
@@ -54,31 +54,38 @@ http {
...
     }
-    # prevent clickjack attacks
-    add_header X-Frame-Options "SAMEORIGIN";
-    # prevent XSS attacks
-    add_header X-XSS-Protection "1; mode=block";
-    # Configuring HSTS
-    add_header Strict-Transport-Security 'max-age=31536000; includeSubDomains; preload';
+
+    # If user agent is a program/bot, block
+    if ($http_user_agent ~* "java|curl|python") {
```
123joshuawu, line 64: do we need/want this? If we're trying to debug something, I can see us using curl, maybe python. probably also not sustainable to block list of languages

src/web/nginx.conf
```diff
@@ -54,31 +54,38 @@ http {
...
+    
+    # Enable session tickets
+    ssl_session_tickets on;
 
     # OCSP stapling
     ssl_stapling on;
     ssl_stapling_verify on;
-    resolver 1.1.1.1 1.0.0.1 [2606:4700:4700::1111] [2606:4700:4700::1001]; # Cloudflare
-
+    resolver 8.8.8.8 8.8.4.4 208.67.222.222 208.67.220.220 valid=60s;
```
123joshuawu, line 87: What is this pointing to?

src/web/nginx.conf
```diff
@@ -90,12 +97,23 @@ http {
     # serve Flask
     location / {
       root /app;
+      proxy_set_header Connection "";
```
123joshuawu, line 100: What's this for?

src/web/src/App.vue
```diff
@@ -5,44 +5,17 @@
 </template>
 
 <script>
-import { TOGGLE_DARK_MODE, SET_COURSE_LIST } from "@/store";
```
123joshuawu, line 8: Are these changes from merging? should maybe prefer rebasing to avoid having these commits duplicated and unexpected changes

src/web/nginx.conf
RichtXO: classic Google and Cisco... should have a comment on that!

src/web/nginx.conf
RichtXO: I think that was my attempt to keep the connection alive, but I forgot the reason why I wanted that...

src/web/nginx.conf
```diff
@@ -54,31 +54,33 @@ http {
...
-    # prevent clickjack attacks
-    add_header X-Frame-Options "SAMEORIGIN";
-    # prevent XSS attacks
-    add_header X-XSS-Protection "1; mode=block";
-    # Configuring HSTS
-    add_header Strict-Transport-Security 'max-age=31536000; includeSubDomains; preload';
+    
+    # Headers
+    add_header X-Frame-Options "SAMEORIGIN";                                                      # prevent clickjack attacks
+    add_header X-XSS-Protection "1; mode=block";                                                  # prevent XSS attacks
```
IkeLyons, line 65: I think this header is only going to work for non-modern browsers. Its not in the scope of this PR, but we should look at other ways to prevents XSS.

IkeLyons (approved):
Looks good to me! lgtm

src/web/nginx.conf
RichtXO: I think in the future, in regards to security, we might want to switch to [bunkerized nginx](https://github.com/bunkerity/bunkerized-nginx). It would automate all these security practices. It also automates the SSL cert too

PR #426, opened 2021-03-17
Updated CSVs

RichtXO:
**Issue**

Just updating the course lists

**Additional Info**

An auto-update feature would be nice 💯


123joshuawu:
Looks good. Probably want to break out the ci fix to a separate pr

Issue #121, opened 2020-04-22
Distribution — RPI : Auto Schedule Update


jshom:
**Job**
As a student, I don't want to schedule with false data. Often times, the schedule often changes under our feet, a class that was once offered at 10-12 on mondays/thur is now on tuesday/friday and we need to be able to adapt to that.

**Describe a possible solution**
In a separate repo YACS-RCOS/yacs-rpi-fountain or a better name, I'm horrible at naming

- [ ] Copy parsing functionality from (yacs)/rpi-data/
- [x] Create github action that:
  - Pulls data from specified sources: i.e list https://sis.rpi.edu/reg/zs20200501.htm (for list of specified semester)
  - Sends a semesters csv to a set of YACS api's (env var's)
  - Runs every 2 hours

**Context**
While dogfooding this to schedule my classes, the class Ontologies changed its scheduling type from 12-12 M/T to 12-4 only on monday, we need to be able to pick this up on the fly. I could have missed registration for a few critical classes

**More Info**
Another intention of this is to fully extract RPI data functionality from the central repo of YACS. RPI is an instance of YACS, not RPI-YACS. This is critical for YACS's growth.

Note this should only update one semester at a time, we can expand to updating multiple semesters at a time, but it does not make much sense as the schedule does not change for last years schedule.

PR #436, opened 2021-03-29
Auto update prod fall 2021 courses every hour

123joshuawu:
Part of #121 

Feel free to change the interval and add/remove semesters as needed

### DIFF
--- a/.github/workflows/prod-update.yaml
+++ b/.github/workflows/prod-update.yaml
@@ -1,0 +1,47 @@
+name: Prod Update
+
+on:
+  workflow_dispatch:
+
+jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/prod-update.yaml
```diff
@@ -0,0 +1,46 @@
+name: Prod Update
```
RichtXO, line 1: I would prefer the name to be `Semester Data Update` or something related

RichtXO (approved):

jshom (approved):

123joshuawu:
Alright here we go

PR #219, opened 2020-10-07
Remove wait for ci from pr-cd

123joshuawu:
**Issue**

N/A

**Database Changes/Migrations**

N/A

**Test Modifications**

N/A

**Test Procedure**

1) create new pull request
2) `ci` and `pr-cd` should both start

**Photos**

N/A

**Additional Info**

N/A

PR #445, opened 2021-04-14
Adding Lighthouse audit

RichtXO:
Was just thinking of automating this audit for front-end changes

**Additional Info**
We should have had this a long time ago :/

The minScore should be changed before merging with the master. The current values are in place for the current website!


123joshuawu:
can add this as a job to the CI workflow

and not sure if we will need data for this to work properly, but if so, can use this e2e workflow as reference https://github.com/YACS-RCOS/yacs.n/pull/251/files

RichtXO:
The only issue I have with this is the link Lighthouse generates. Like why does it have `lighthouse***nfrastructure.appspot.com` when it's `lighthouse-infrastructure.appspot.com`? Is this like an API limiter? I got no idea, but everything else works as intended!

jshom:
This is awesome!

Finding this in the logs, do we want to fail the test if below the threshold? or do we want to always succeed?
```
 ⚠️  categories.performance warning for minScore assertion
        expected: >=0.8
           found: 0.11
      all values: 0.08, 0.11, 0.1
```

src/web/nginx.conf
```diff
@@ -13,16 +13,14 @@ http {
   default_type application/octet-stream;
   keepalive_timeout  65;
 
-# Expires map
+  # Expires map
```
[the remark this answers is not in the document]
RichtXO: Ok, I think there's something wrong with my machine... mb
I did `git rebase origin/master` but every time I push, it states a merge :/

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,147 @@
...
+        with:
+          node-version: 15.x
+
+      - name: Build and Run Production
+        run: |
+          docker-compose \
+            -f docker-compose.yml \
+            -f docker-compose.production.yml \
+            build &&
+          docker-compose \
```
123joshuawu, line 72: up will build

src/web/nginx.conf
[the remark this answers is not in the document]
123joshuawu: prolly need to force push if you've already merged in the past, unless there are rebase conflicts.
just a suggestion, won't work everytime

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,147 @@
...
+        working-directory: rpi_data
+        run: |
+          curl --insecure -X POST -H "Content-Type: multipart/form-data" \
+          -F "isPubliclyVisible=on" -F "file=@$FILE_NAME" $AUTH_ARGS $API_ENDPOINT
+        env:
+          FILE_NAME: fall-2021.csv
+          AUTH_ARGS: ${{ secrets.AUTH_ARGS }}
+          API_ENDPOINT: https://localhost/api/bulkCourseUpload
+
+      - name: run Lighthouse CI
```
123joshuawu, line 88: could also explore using a lighthouse github action like https://github.com/marketplace/actions/lighthouse-check or https://github.com/marketplace/actions/lighthouse-ci-action

RichtXO: Will definitely take a look at that

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,147 @@
...
+        run: |
+          npm install -g @lhci/cli@0.7.x &&
+          lhci autorun
+        env:
+          LHCI_GITHUB_APP_TOKEN: ${{secrets.LHCI_GITHUB_APP_TOKEN}}
+
+  publish-action-status:
```
123joshuawu, line 95: probably won't need this if pr-cd is in the same file

i believe we also had pr-cd starting at the same time as the tests to speed up deployments but this is fine as well

src/web/nginx.conf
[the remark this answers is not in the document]
123joshuawu: EDIT if you've already merged probably won't do anything i think

.github/workflows/ci.yaml
```diff
@@ -1,77 +0,0 @@
-name: CI
-
-on:
-  push:
```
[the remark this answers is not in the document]
RichtXO: Oh crap.
I was trying to find ways to not deploy the `temp_deploy` if checks failed. Maybe it's best to separate it.

RichtXO:
> This is awesome!!
> 
> Finding this in the logs, do we want to fail the test if below the threshold? or do we want to always succeed?
> 
> https://github.com/YACS-RCOS/yacs.n/runs/2358958949#step:6:33
> 
> ```
>  ⚠️  categories.performance warning for minScore assertion
>         expected: >=0.8
>            found: 0.11
>       all values: 0.08, 0.11, 0.1
> ```

We can definitely change `warn` to `error`, and the values for `minScore`. But it seems ever since PWA was introduced, it's not looking good lol. I guess it's best to set a concrete value and merge as admin. I only let that pass just to have green checks.

.github/workflows/pr.yaml
RichtXO: > Oh crap.
> I was trying to find ways to not deploy the `temp_deploy` if checks failed. Maybe it's best to separate it.

Ye, as explained here lol. That's why I change the `on` event triggers.

.github/workflows/ci.yaml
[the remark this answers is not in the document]
123joshuawu: I think this is what you're looking for? https://github.com/YACS-RCOS/yacs.n/pull/219

.github/workflows/pr.yaml
RichtXO: Looked into them, and they just take info from the `lighthouserc.js` and integrate them into the yaml file. I think it would be better to specify the specs and other lighthouse-related things in a separate file.

123joshuawu (approved):

jshom (approved):

PR #447, opened 2021-04-16
Image export bug

### LINKED ISSUE #420: Bug — Export to Image not working
**Describe the bug**
The schedule view image export does nothing.

**To Reproduce**
1. Go to YACS
2. Add some courses
3. Click on the 'Export data' dropdown
4. Click on 'Export to image'
5. Notice that no image is downloaded

**Expected behavior**
Schedule image should export as added in #339 

**Additional context**
This problem persists both logged out and logged in.

keerti-sundaram:
**Issue**

Note which issue number(s) this Pull Request closes. All pull requests should close an issue, if one does not exists, make one. This helps us plan and discuss before we decide on implementation

*Example:*
> closes #420 

Updated CourseScheduler.vue to set id after user has selected which subsemester. This way, the ids do not overlap in the case of sub semesters.

**Test Procedure**

*Example:*
> 1. Go to YACS
> 2. Select a subsemester (ex: SUMMER 2021)
> 3. Select date range from drop down, add courses
> 4. click export to image, image is downloaded


RichtXO:
Good work! Maybe in the future, we could find a way to make it download both semesters regardless what semester calandar user has selected. 👍 
Also, let's fix the conflicting file, then merge it in!

RichtXO (approved):

jshom (approved):

PR #448, opened 2021-04-19
Responsive schedule

### LINKED ISSUE #444: Feature Request — Schedule view height size responsive
**Job**
Make schedule size responsive to screen size.

Examples Issue Screenshots:

Screen Shot 2021-04-14 at 12 30 08 AM

Screen Shot 2021-04-14 at 12 29 33 AM

**Proposed Solution**

Keep height responsive on resize to 80% of the height for all devices with a min height of 600px or so (what it is right now)

Viyerelu23333:
**Issue**
> closes #444

Untouched.

**Test Modifications**
Untouched.

**Test Procedure**
1. Import some courses
2. Add some courses into schedule
3. Change the height of the viewport, the positions of courses should remain the same
4. If the schedule's height is less than 600px, the schedule's size should remain the same

**Photos**
Firefox
FF_Before
FF_After
Blink/Webkit
Webkit_After

**Additional Info**
Compatibility: [CSS Math Function](https://caniuse.com/css-math-functions)

**Some Notes**
- ~~Do we really need the `overflow-y: auto` in `ScheduleEvent: event-text`, i.e. the scroll bar on every course?~~
- This may cause lags when constantly changing the height of the viewport.
- Minimized the scroll bars in the schedule.

### DIFF
--- a/src/web/src/App.vue
+++ b/src/web/src/App.vue
@@ -56,6 +56,15 @@ export default {
       );
     },
   },
+  computed: {
+    colorSchemeMeta() {
+      return this.isFollowingDeviceColor()
+              ? "light dark"
+              : this.$cookies.get(COOKIE_DARK_MODE) === "true"
+                  ? "dark"
+                  : "light";
+    },
+  },
   metaInfo() {
     return {
       title: "YACS",
@@ -75,6 +84,7 @@ export default {
         { property: "og:site_name", content: "YACS" },
         { property: "og:type", content: "website" },
         { name: "robots", content: "index" },
+        { name: "color-scheme", content: this.colorSchemeMeta },
       ],
     };
   },
--- a/src/web/src/components/Schedule.vue
+++ b/src/web/src/components/Schedule.vue
@@ -43,7 +43,7 @@
             borderColor: getBorderColor(courseSession),
             color: getTextColor(courseSession),
             width: dayWidth + '%',
-            overflow: hidden,
+            overflow: 'hidden',
           }"
         ></ScheduleEvent>
         <div
@@ -97,7 +97,7 @@ export default {
     getTextColor,
     /**
      * Calculate the height of the schedule block for `courseSession`
-     * Returns the px height
+     * Returns the vh height
      * @param {CourseSession} courseSession
      * @returns {number}
      */
@@ -119,7 +119,7 @@ export default {
     },
     /**
      * Calculate the position of the schedule block for `courseSession`
-     * Returns the px margin-top
+     * Returns the vh margin-top
      * @param {CourseSession} courseSession
      * @returns {number}
      */
--- a/src/web/src/components/Schedule.vue
+++ b/src/web/src/components/Schedule.vue
@@ -1,7 +1,7 @@
 <template>
   <div
     class="schedule"
-    :style="{ height: totalHeight + 'px' }"
+    :style="{ height: totalVHeight + 'vh', 'min-height': minHeight + 'px' }"
     data-cy="schedule"
   >
     <div class="schedule-legend">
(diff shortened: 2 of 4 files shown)


marchdan:
@Viyerelu23333 There are 3/4 options for the course blocks:
> 1. Scroll bars on blocks that have more data than can visibly fit in the window
> 2. The Course blocks expand when you hover over them so you can see all of their content
> 3. You can click on a course block and a pop-up comes up with all of the schedule relevant info for that section of the course
> 4. Noting happens and you see what you see.

I personally think option 3 would fit our needs and aesthetic the best atm. Number 1 is what we have right now, and number 2 is a feature that was being worked on for old YACS.

Viyerelu23333:
> @Viyerelu23333 There are 3/4 options for the course blocks:
> 
> > 1. Scroll bars on blocks that have more data than can visibly fit in the window
> > 2. The Course blocks expand when you hover over them so you can see all of their content
> > 3. You can click on a course block and a pop-up comes up with all of the schedule relevant info for that section of the course
> > 4. Noting happens and you see what you see.
> 
> I personally think option 3 would fit our needs and aesthetic the best atm. Number 1 is what we have right now, and number 2 is a feature that was being worked on for old YACS.

Oh I see.  Totally agree that we might need option 3.  I tried to add a `hover` as option 2, and this is not friendly to mobile users.

jshom:
> > @Viyerelu23333 There are 3/4 options for the course blocks:
> > > 1. Scroll bars on blocks that have more data than can visibly fit in the window
> > > 2. The Course blocks expand when you hover over them so you can see all of their content
> > > 3. You can click on a course block and a pop-up comes up with all of the schedule relevant info for that section of the course
> > > 4. Noting happens and you see what you see.
> > 
> > 
> > I personally think option 3 would fit our needs and aesthetic the best atm. Number 1 is what we have right now, and number 2 is a feature that was being worked on for old YACS.
> 
> Oh I see. Totally agree that we might need option 3. I tried to add a `hover` as option 2, and this is not friendly to mobile users.

Interesting, gonna make an another issue for this to investigate. For the issue (#444) however, I think we can leave the scroll bars and more info the same as it is and handle that separately.

src/web/src/components/Schedule.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div
     class="schedule"
-    :style="{ height: totalHeight + 'px' }"
+    :style="{ height: totalVHeight + 'vh', 'min-height': minHeight + 'px' }"
```
jshom, line 4: smol nit, perhaps we can use the js template strings like so:
```diff
- :style="{ height: totalVHeight + 'vh', 'min-height': minHeight + 'px' }"
+ :style="{ height: `${totalVHeight}vh`, 'min-height': `${minHeight}px` }"
```
wyt? (also can be tweaked @ [lines 39-42](https://github.com/YACS-RCOS/yacs.n/pull/448/files#diff-3b36822eec87cadead13937bf56766759546ea4f5eaca69b8e3aa3caeae3552eR39-R42))

yangytjason2000 (commented):
It's good to have this feature for YACS. The code looks good.

jshom (approved):

marchdan (approved):
I wish the scroll bars weren't so thick, but that's not a problem for this issue

marchdan:
@Viyerelu23333 when the screen is small on desktop the horizontal scrollbars are still thick-bois

src/web/src/App.vue
```diff
@@ -56,6 +56,15 @@ export default {
       );
     },
   },
+  computed: {
+    colorSchemeMeta() {
+      return this.isFollowingDeviceColor()
+              ? "light dark"
+              : this.$cookies.get(COOKIE_DARK_MODE) === "true"
```
marchdan, line 63: Is this the current way we are checking for dark mode? I remember that we were trying to move away from putting things like this in cookies.

marchdan (commented):
Looks good minus possibly changing how we check dark mode

src/web/src/App.vue
Viyerelu23333: I think this part does not involve dark mode checking...  
The `"light dark"` string is for `<meta>` tag.  It enables the UA to choose the correct UA style for light/dark mode, i.e. the light/dark scrollbars in this case.  

The dark mode decision is at L29-L58 which is not listed here.

marchdan: > I think this part does not involve dark mode checking...
> The `"light dark"` string is for `<meta>` tag. It enables the UA to choose the correct UA style for light/dark mode, i.e. the light/dark scrollbars in this case.
> 
> The dark mode decision is at L29-L58 which is not listed here.

I saw the cookie check so I wanted to be sure since I know we were moving aware from storing things that we need in order to run properly in the cookies

jshom (approved):

PR #453, opened 2021-04-30
Adjusting margins

NickKing24:
Added a margin to the color blind assist toggle so it doesn't touch the schedule grid and the left side is in line with the export data button. Also added a margin to the CRNs so they also don't touch the grid.

**Photos**

Before:
image

After:
image


RichtXO (approved):
GJ mate!

jshom (approved):

PR #471, opened 2021-07-23
Fixed scrollbar darkmode

### LINKED ISSUE #465: Bug — Scroll Bars Don't Change Colors
**Describe the bug**
The scroll bars don't immediately change colors when switching between dark and light mode.

**To Reproduce**
1. Go to YACS
2. Note that the scroll bar follows the current mode's color scheme
3. Toggle to the opposite color mode
4. Scroll bar remains in the original color
5. Refresh the page
6. Scroll bar now follows the new mode 

**Expected behavior**
Scroll bar should immediately change when toggling modes

antmcgrann:
closes #465 

Fixes issue of scrollbar not changing colors after you switch to/from dark mode

### DIFF
--- a/rpi_data/modules/fetch_catalog_course_info.py
+++ b/rpi_data/modules/fetch_catalog_course_info.py
@@ -1,14 +1,14 @@
 import requests as req
-import threading
+import threading #https://docs.python.org/3/library/threading.html
 import unicodedata
 import re
-import regex
-import json
+import regex #https://www.dataquest.io/blog/regex-cheatsheet/
+import json 
 from datetime import date
 from time import time
 from threading import Lock
 from io import StringIO
-from bs4 import BeautifulSoup, SoupStrainer
+from bs4 import BeautifulSoup, SoupStrainer #https://www.crummy.com/software/BeautifulSoup/bs4/doc/
 from lxml import etree
 
 chunk_size = 200 # max number of course ids per GET request
@@ -273,6 +273,7 @@ class acalog_client():
 
 def main():
     c = acalog_client(acalog_api_key)
+    print(acalog_api_key)
     courses = c.get_all_courses()
     if dev_output_files:
         dwrite_utf8_file(json.dumps(courses, indent=4), "courses21.json")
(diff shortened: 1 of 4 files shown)


rpi_data/modules/fetch_catalog_course_info.py
```diff
@@ -273,6 +273,7 @@ def get_all_courses(self):
 
 def main():
     c = acalog_client(acalog_api_key)
+    print(acalog_api_key)
```
marchdan, line 276: Was this print left in on purpose?

marchdan (changes requested):
Using page reloads to fix this issue is bulky and does not present well on the temp deploy. You might look at how other sections of the site handle the dark mode toggle for everything without having to do a page reload.

Viyerelu23333:
Not sure if refreshing on color change is the best way to solve it...  
And your refresh did not catch the media queries changes.

Firefox does provide a good solution to this by detecting the background of the box automatically.  Safari overwrites the UA styles, so the scrollbar is transparent and color-aware.  However, Chrome will only detect the `color-scheme` meta tag.  

Admittedly, the `color-scheme` can be `light dark` to specify both modes.  But in stylesheet, we are not using media queries to detect the dark mode, so Chrome cannot detect what mode is it currently and fallbacks to `light`.  Then the scrollbars are in light.

A possible fix to this bug is to make the `color-scheme` dynamically or change the stylesheets...

marchdan:
@Viyerelu23333 The scroll bar wouldn't have a `fill= "currentcolor"' property, would it?

Viyerelu23333:
> @Viyerelu23333 The scroll bar wouldn't have a `fill= "currentcolor"` property, would it?

There are CSS selectors for that.  In WebKit, use `::-webkit-scrollbar-thumb { background-color }` and `::-webkit-scrollbar-track { background-color }`.  But this breaks the UA styles.

EthanTuttle:
Developer's fix wasn't accepted and is no longer active, closing this branch and pull request. Someone else please take over the issue (maybe @SamooMei, he's pretty good at this stuff)

PR #472, opened 2021-08-10
Simplified Column Logic

### LINKED ISSUE #295: Tech Debt — Simplify Explore Page
In https://github.com/YACS-RCOS/yacs.n/blob/master/src/web/src/pages/CourseExplorer.vue

The logic to create two columns is relatively difficult to maintain and understand, convert grid to use `<b-col md="6">` elements within a `<b-row>`. This will eliminate creating separate columns and improve mobile usability

NickKing24:
**Issue**

> closes #295

**Additional Info**

Not fully sure if this what the issue is asking for, but added a v-for loop in the column logic so the code to create a column isn't duplicated anymore.


EthanTuttle (approved):
Extra nesting was removed and functionality is still there so I think this solves the issue

PR #478, opened 2021-09-22
Remove pr-cd workflows

123joshuawu:


RichtXO:
Do we really want to fix this? lol

123joshuawu:
either we fix it or take it out. It's failing all our checks

RichtXO:
I think taking out would be better, unless Jacob wants to meet us sometime in the foreseeable future :)

RichtXO (approved):
We still gonna pay Jacob a visit :)

Issue #449, opened 2021-04-21
Discussion (Pre Feature Request) — Schedule events scroll bar and info


jshom:
**Job**
Discuss and propose changes to event info based on sizing/device differences

Moving discussion from #448 about event info and scroll bars here.

---

> > @Viyerelu23333 There are 3/4 options for the course blocks:
> > > 1. Scroll bars on blocks that have more data than can visibly fit in the window
> > > 2. The Course blocks expand when you hover over them so you can see all of their content
> > > 3. You can click on a course block and a pop-up comes up with all of the schedule relevant info for that section of the course
> > > 4. Noting happens and you see what you see.
> > 
> > 
> > I personally think option 3 would fit our needs and aesthetic the best atm. Number 1 is what we have right now, and number 2 is a feature that was being worked on for old YACS.
> 
> Oh I see. Totally agree that we might need option 3. I tried to add a `hover` as option 2, and this is not friendly to mobile users.

Interesting, gonna make an another issue for this to investigate. For the issue (#444) however, I think we can leave the scroll bars and more info the same as it is and handle that separately.

marchdan:
The original issue was being able to screenshot/export your YACS schedule and see all relevant info if possible, so the following items need to be top priority visible:
```
Course Label (I.e. CSCI 1100)
Course Type (Lcture/Lab/Test/etc.)
Location (DCC 308)
```
If the screen is large enough to fit a course name it can be whatever fits height-wise of:
```
Course Name (Computer Science 1)
Course Label (I.e. CSCI 1100)
Course Type (Lecture/Lab/Test/etc.)
Location (DCC 308)
Instructors (Turner)
CRN (123456)
```

Beyond that, I think the user would be ok clicking on a course block to see additional info. Option 3 is equally friendly to mobile and desktop use of said feature. Scroll bars are ok on mobile, but a little unwieldy atm and hover elements will not behave nicely with mobile.

123joshuawu:
Linking some previous thoughts in #227

PR #470, opened 2021-07-23
Professor names + emails on individual course pages

i-am-david-liu:
**Issue**
> Closes #381

Note: doesn't fully close #381, as it will only add professor emails and their names, the latter which is unrelated to the issue entirely. Should I make a new issue that's more specific?

**Database Changes/Migrations**
> Adds "course_instructor" and "email" column to .csv

**Photos**

to

**TO DO**
- Display emails for courses with more than one instructor
- Revise test cases

PR #425, opened 2021-03-15
Moving to vuex

123joshuawu:
**Issue**

Closes #419 
Closes #404 
Closes #384 
Closes #343 

**Database Changes/Migrations**

N/A

**Test Modifications**

N/A

**Test Procedure**

Do anything and nothing should break

**Photos**

N/A

**Additional Info**
This PR is to consolidate the scattered state variables into vuex. So we're talking `subsemesters`, `courses`, `departments`, `semesters`. In the future, the student course selection and some other things will need to be moved as well but we'll save that for later. 

This should fix some of the issues arising from state conflicts. 

One thing to note is that semester changes no longer go through the router, it is entirely done within vuex. This essentially deprecates the use of the query string semester option for CourseScheduler. I've left the functionality in for now for backwards compatibility, it will work on initial load, but any subsequent semester changes will not be reflected by the query string. 

The benefit to this is all components are able to access the same `selectedSemester` state. This means Explore and CourseScheduler pages are all utilizing the same `selectedSemester` and `courses` state. So any semester changes will update both accordingly. 

One final note is the use of the navigation guard for `Student` routes to load the default semester. On initial load, because the user is navigating to the page for the first time, while the navigation guard is executing, the user will be presented with a blank page. Ideally, the navigation guard logic should run almost immediately and not be erroneous. However, the best solution would be to implement 1) a loading screen for the initial navigation guard and 2) an error screen in case the initial navigation guard fails.

PR #481, opened 2021-09-25
More Info in Schedule Events

Viyerelu23333:
**Issue**
Related to #449

Added `session_type` and `instructor` to `course_session`.

**Test Modifications**
No.  Columns in `course_session` table do not have any unit test.

**Test Procedure**
- Add some courses, and there will be simplified information in the scheduler (course code, session type and location).
- Click on the schedule event should show the detailed information (course name, instructor, time period, CRN, etc.).
- Click on the course link should redirect to the correct course with correct selected semester.

**Photos**
PC

**Additional Info**
- Feel free to discuss the approaches of displaying overflowed info in Schedule Events.
- ~~The popover cannot be self-dismissed when clicking to other schedule events, maybe modal should be used here?~~ Fixed
- ~~How to let users know that the schedule events are actually clickable?~~ Moved to `hover`
- Now `hsl()` is implemented in `ColorService`, so colors could reduce saturation and luminosity in dark mode.
- The height of scheduler changed to `80vh` to fill the gap at the bottom.
- Added a new cookie `selectedSemester` to hold the state between refreshes.


EthanTuttle:
Maybe the class type should go lower? Based on that issue #449 we should keep the higher priority information towards the top. I'm not sure if the class type is as important as some of the things below it like professor

marchdan:
@RichtXO Do we have temp deploy's disabled? I'd love to review this, but there is no temp deploy

RichtXO:
@marchdan ye unfortunately. I'm still in a process of replacing the temp deploy with a staging server. Almost done with that!

marchdan:
Also, how does the work done here effect #470?

Viyerelu23333:
> Also, how does the work done here effect #470?

@marchdan They are compatible except one of the branches needs to update to a newer alembic migration, depending on which comes first to the master branch...  I guess.

marchdan:
> > Also, how does the work done here effect #470?
> 
> 
> 
> @marchdan They are compatible except one of the branches needs to update to a newer alembic migration, depending on which comes first to the master branch...  I guess.

I'd say since #470 seems to be a stale PR atm, I'd say that once this PR is completed and merged, someone should update/fix 470 as needed if not finish it.

Viyerelu23333:
> > Well... fixed in a brutal-force way.  Back button still returns to the default semester :)
> 
> Strange, when someone navigates to the regular site, it should automatically navigate to selected semester... (Unless that's now how it's implemented and refreshing goes back to default semester @RichtXO ?)

@marchdan 
I think there was a cookie that stores the current semester...  Might be related to #425?

marchdan:
> > > Well... fixed in a brutal-force way.  Back button still returns to the default semester :)
> 
> > 
> 
> > Strange, when someone navigates to the regular site, it should automatically navigate to selected semester... (Unless that's now how it's implemented and refreshing goes back to default semester @RichtXO ?)
> 
> 
> 
> @marchdan 
> 
> I think there was a cookie that stores the current semester...  Might be related to #425?  

That's how I originally implemented it I think, but I thought it was changed since then? If it's in a cookie than I would think it should be a rather easy fix?

marchdan:
> @marchdan @RichtXO Well... fixed in a brutal-force way. Back button still returns to the default semester :)

This is still happening, but things are otherwise good. Are we approving and making a bug issue for this? Or do we have a fix for it?

marchdan:
@RichtXO @Viyerelu23333 I would love to get this fully approved so we can fully check the UI of #506, any thoughts on the issue of going backward in Explorer? (Issue also occurs if you navigate on a pre-req link FYI)

Viyerelu23333:
> I would love to get this approved so we can merge and fully check the UI of #506, any thoughts on the issue of going backward in Explorer? (Issue also occurs if you navigate on a pre-req link FYI)

@marchdan 
This issue deserves a better solution, just like before.  How about reverting the last two commits and having a new issue of it?

marchdan:
@Viyerelu23333 I still need to go in and properly test this PR in its current state. But if y'all are looking to have these finished by the weekend, the current state of the PR works for the relevant functionality. You can get to the right course from the schedule. It would be nice to have prerequisite links work too, but it's better to have it partly working than not working.

Viyerelu23333:
> I still need to go in and properly test this PR in its current state. But if y'all are looking to have these finished by the weekend, the current state of the PR works for the relevant functionality. You can get to the right course from the schedule. It would be nice to have prerequisite links work too, but it's better to have it partly working than not working.

@marchdan  Sorry if I am wrong, I have no experience to the VueX...
It seems the states from VueX cannot be preserved when refreshing the page. And I am not sure if the backend API could read the VueX states.  We probably need a method to store the `selectedSemester` into some persistent storage like local (session) storage or cookies.

marchdan:
> > I still need to go in and properly test this PR in its current state. But if y'all are looking to have these finished by the weekend, the current state of the PR works for the relevant functionality. You can get to the right course from the schedule. It would be nice to have prerequisite links work too, but it's better to have it partly working than not working.
> 
> 
> 
> @marchdan  Sorry if I am wrong, I have no experience to the VueX...
> 
> It seems the states from VueX cannot be preserved when refreshing the page. And I am not sure if the backend API could read the VueX states.  We probably need a method to store the `selectedSemester` into some persistent storage like local (session) storage or cookies.  

I am still confused as to when it was removed from cookies, lol.

Viyerelu23333:
@marchdan 
Now the cookie is back :)

RichtXO (approved):

PR #486, opened 2021-10-01
Update course data in repo automatically

### LINKED ISSUE #121: Distribution — RPI : Auto Schedule Update

RichtXO:
**New Feature**

Now it should automatically update not only the production site but also the repository! 

> Closes #121 

**Additional Info**

Why wasn't this implemented before?????

### DIFF
--- a/.github/workflows/prod-update.yaml
+++ b/.github/workflows/prod-update.yaml
@@ -49,4 +49,4 @@ jobs:
           FILE_NAME: ${{ matrix.semester }}.csv
           AUTH_ARGS: ${{ secrets.AUTH_ARGS }}
           API_ENDPOINT: ${{ secrets.UPDATE_API_ENDPOINT }}
-        working-directory: rpi_data    
+        working-directory: rpi_data
--- a/.github/workflows/prod-update.yaml
+++ b/.github/workflows/prod-update.yaml
@@ -16,6 +16,9 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
+        with:
+          persist-credientials: false
+          ref: ${{ github.head_ref }}
 
       - uses: actions/setup-python@v2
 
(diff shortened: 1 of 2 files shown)


123joshuawu (commented):
Looks good. But we do need to note that this will add a commit to the master branch every hour. I think we also determined that it wasn't really necessary for developers to have the most up to date hourly course data. Probably could setup a separate workflow, but probably would want to have this happen with a longer interval.

RichtXO:
True, I do agree devs don't really need up-to-date data for testing, but it would be nice to have.
And yea, having a commit every hour wouldn't be ideal, probably every month would do.

RichtXO:
I also realized the links to get the data don't stay active forever... maybe it's best to not have it...

RichtXO:
@marchdan Yep! And I did make the job to update the repo once a month!

marchdan:
Question: Does any of this automatically update the course data on prod?

RichtXO:
@marchdan Yep! But the link may not be active in a couple of months because RPI closes them. Oh, I don't know if it would push an empty CVS file tho when the link is not active.

PR #493, opened 2021-10-09
Resolved Default Semester Bug

### LINKED ISSUE #489: Default Semester Bug — Not changing default semester to desired one
**Describe the bug**
Currently, how YACS.N saves the default semester is to put it into a table and references it. When the admin wants to change the default semester to something else, the API just adds the desired semester into the table and does not remove the old one.

**To Reproduce**
1. Load 2 semester data. The first semester uploaded should be the default semester
2. Go into Admin panel, change default semester to the 2nd uploaded semester
3. YACS still has the first semester as the default semester

**Expected behavior**
When admin changes the desired default semester, the website would update to show the semester by default.

**Additional context**
I think it's as easy as adding a `DELETE FROM ... ` line before adding the desired semester. :smile:

Viyerelu23333:
**Issue**
Closes #489

Will delete any previous default semesters before adding/update current default semester (`admin_settings.semester`).
No new migrations.

**Test Modifications**
Not yet.

**Test Procedure**
> 1. Load 2 semester data. The first semester uploaded should be the default semester
> 2. Go into Admin panel, change default semester to the 2nd uploaded semester
> 3. ~~YACS still has the first semester as the default semester~~
3. YACS now should correctly update the default semester

**Photos**
None ( ° ∀ ｡ ).

**Additional Info**
The cache for `GET /api/defaultsemester` is disabled.


Viyerelu23333:
> > Yep! That's probably because we have 2 caching subsystems going on (one from last, another from NGINX). Flask has to finish building its cache first, then NGINX builds its own cache, then it would be delivered to the end-user lol. We are reworking the NGINX with the [GH-499](https://github.com/YACS-RCOS/yacs.n/pull/499).
> 
> So once #499 is merged master can be pulled into here to check the fix and then merge this?

@marchdan 
Sorry for the late reply.  I merged `master` into this branch.

marchdan (approved):
LGTM when testing on local dev script.

Issue #504, opened 2021-10-28
Bug — Limited Size of Cookies


Viyerelu23333:
**Describe the bug**
Since browsers have a 4096-byte limit for each cookie, if a user selected many courses that exceed the limit, courses will not be loaded after refreshing.

**To Reproduce**
- Log out.
- Select any 6 courses in Summer 2021
- Select any 6 courses in Fall 2021
- Select any courses in Spring 2022
- Refresh the page
- Some/All selected courses in Spring 2022 are missing.

**Expected behavior**
The selected courses should be displayed.

**Screenshots**
![Out of Cookies](https://user-images.githubusercontent.com/11363469/139331054-9834ff2c-07f1-451f-9bf1-d8ea86537c84.png)

**Additional context**
It could be solved by optimizing the data structure of cookies, or switching to local storage, ~~or deleting the old selected courses when reaching the limit~~.
However, this is not a significant issue if the cookie expires before the new registration.

Viyerelu23333:
> I am assuming you found this while logged out? Course storage while logged in should be in the database not cookies so this shouldn't be a problem there.
> 
> More on topic, EDIT: I believe we discussed when courses were added back to cookies that there should be a better way to do it, but we weren't sure how. I believe we are already storing them efficiently (unless you wanted to have each semester's courses be it's own cookie 🤔). The solution would either be to limit the number of courses that can be stored overall and prompt users to sign-in once they reach it. Or find a better way to store logged-out saved courses.

Oh, I did not log in...  The `selectedCourses` data structure contains `_getCourseIdentifier()` from `YacsService`, which returns a relatively large chunk of data...  
I am still looking at the mechanisms of loading courses.

PR #506, opened 2021-10-29
All possible schedules

### LINKED ISSUE #52: Scheduling — Show all possible schedules based on selected sections
Instead of notifying users of a schedule conflict when adding a course section, indicate conflicting course sections in selected courses, then when a user selects a conflicting course section, generate all combinations of course sections that do not have a schedule conflict and their corresponding schedule. 

I'm thinking redesign of Schedule to an immutable class created using the factory design pattern.
Make Schedules lightweight and hashable to provide fast equality comparisons.

101polarbears:
**Issue**

closes #52 

**Additional Info**

some know issues: the cookies are not yet available (refreshing the page does not keep your changes), Nick is working on it
some aesthetic stuff to be reviewed (the current color may be identical for different courses)


marchdan:
@NickKing24 Would your most recent commit close #504 ?

NickKing24:
> @NickKing24 Would your most recent commit close #504 ?

No I don't think so because I just reimplemented the old cookie.

123joshuawu:
Very nice, is this done done? or just done

marchdan:
I found some issues of various caliber, will list them here as I find more:
- Course block scroll doesn't work (Won't matter once #481 is merged, but good to be aware of)
- Clicking on a course that can't fit still adds the course. (Do we want this behavior? Or should it throw an error like it did before and not add it?)
-

NickKing24:
> * Changing between semesters doesn't clear out the old semester from the scheduler, so there is a semester selection bar above the schedule.

This should be fixed now.

src/web/src/pages/NewCourseScheduler.vue
```diff
@@ -473,6 +466,80 @@ export default {
...
+          }];
+        }
+        const result = this.generateSchedule(Object.values(this.selectedCourses))
+        if (!result.length) {
+          throw new Error('conflict!')
+        }
+        this.possibilities = result
+        
+        //Don't set this.index to 0 if just loaded cookie
+        if (this.loadedIndexCookie == 2) {
```
123joshuawu, line 485: if we're generating the new schedules here, shouldn't we always reset the index to 0?

src/web/src/pages/NewCourseScheduler.vue
```diff
@@ -62,17 +62,38 @@
...
+            <b-col class="m-2">
+              <b-button
+                @click="changeSchedule(-1); updateIndexCookie();"
+                size="sm"
+              >
+                Prev
+              </b-button>
+            </b-col>
+            <b-col cols="8" class="m-2 text-center">
+              <span v-if="scheduleDisplayMessage === 2">
```
123joshuawu, line 76: `if (possibilities.length === 1 && possibilities[0].conflict) ... else if (possibilities.length === 1) ... else`

src/web/src/components/Schedule.vue
```diff
@@ -81,8 +77,8 @@ export default {
     ScheduleEvent: ScheduleEventComponent,
   },
   props: {
-    schedule: {
-      default: () => new Schedule(),
+    possibility: {
```
123joshuawu, line 80: why are we renaming this to `possibility`? a possibility is a schedule but a schedule is not necessarily a possibility

src/web/src/components/Schedule.vue
```diff
@@ -93,6 +89,7 @@ export default {
       endTime: 1320,
       totalVHeight: 70,
       minHeight: 600,
+      temp: this.possibility
```
123joshuawu, line 92: what is this for?

123joshuawu (commented):
some initial thoughts

src/web/src/pages/NewCourseScheduler.vue
NickKing24: If we just always set the index to 0, when the index cookie is loaded, it will load the saved index but then set it back to 0 so the index won't be the one from the cookie

src/web/src/pages/NewCourseScheduler.vue
```diff
@@ -473,6 +466,80 @@ export default {
...
+        this.index = l - 1
+        return
+      }
+      if (c >= l) {
+        this.index = 0
+        return;
+      }
+      this.index = c
+    },
+    updateIndexCookie() {
```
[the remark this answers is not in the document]
NickKing24: We also need to update the cookie when new schedules are generated so I thought having it in a method would be better.

marchdan (commented):
As far as current functionality goes, my notes are below:
- It would be nice to somehow mark the sections of a course that don't work with any possible schedule
- If there is any efficient way to mark the courses that won't work in the schedule, that would be amazing to have, but the way that works rn is a good MVP.
- Export to image does not seem to work at all, no file shows up (not sure if this is a new issue )
- Core functionality looks good overall, I am excited to see how it looks when this PR and the more info in schedule events PR are both in master <3

src/web/src/services/YacsService.js
```diff
@@ -40,18 +40,36 @@ const _getCourseIdentifier = (courseObj) => {
   `;
 };
 
+export const newGetCourses = (semester, search = null) => client.get("/class", { params: { semester, search } });
```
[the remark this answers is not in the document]
101polarbears: No, we could remove that

NickKing24:
> * Export to image does not seem to work at all, no file shows up (not sure if this is a new issue )

Yeah it was a new issue. Should be fixed now along with export to ICS which also wasn't working properly.

Viyerelu23333:
It seems the conflict checker failed on sub-semesters.  ~~Or I did something wrong...~~

To reproduce:
- Load `summer-2021.csv`
- Filter Sub-semester `5/24-7/2`
- Select `CHEM-2230`
- Filter Sub-semester `7/12-8/20`
- Select `ARTS-4080`
- The scheduler indicates conflict on both sub-semesters.

RichtXO:
@Viyerelu23333 Good point that you mentioned about the Summer semester. But I would say that this is good to go. Since we are rewriting from scratch anyway, it's good for people to see this feature. We will definitely get every done by the end of this semester :)

RichtXO:
Also, this was suppose to be merged before Fall registration started :/

Issue #496, opened 2021-10-13
Bug — Website refreshes when updated


RichtXO:
**Describe the bug**
Every time we update the prod environment, the client would experience a constant website refresh, not being able to use the website

**Expected behavior**
Service worker is behind this bug. Service worker should have updated itself and remove its old identity stored on the client-side. 

**Screenshots**
_Will update when coming across this error again._

**Additional context**
PWA is most likely misconfigured.

PR #509, opened 2021-11-09
#496-SEUDT

SamooMei:
**Issue**
Richard's Magnum Opus (PWA) is broken in production, so I deleted it.

**Test Procedure**

1. Get on local server
2. Check if you can install the PWA


RichtXO:
This removes the PWA, but not the one that exists on clients. The following repo could be the solution.
https://github.com/NekR/self-destroying-sw/tree/master/packages/webpack-remove-serviceworker-plugin

PR #537, opened 2022-02-16
Fast api course and event

yliao-dev:
**Issue**

Wrote test cases for api/user/course.
Created file "test_user_course.py"

**Test Modifications**
> Added test `tests/api/db/test_user_course.test_user_course_post_success` to ensure all the post API calls can post sucessfully
> Added test `tests/api/db/test_user_course.test_user_course_post_failure` to ensure all the post API calls fails with incorrect json parameters

**Test Procedure**
Run the single test file from the docker container yacs_api:
pytest -s tests/test_user_course.py

### DIFF
--- a/src/api/tests/test_user_course.py
+++ b/src/api/tests/test_user_course.py
@@ -27,6 +27,7 @@ def test_user_course_post_success(post_user, client: Client):
     assert get_data['crn'] == TEST_USER_COURSE['cid']
     d = client.delete("/api/session", json={"sessionID": sess.json()['content']['sessionID']})
     assert d.status_code == 200
+
 def test_user_course_post_failure(client: Client):
     '''
     Test user course post with invalid parameter
@@ -38,3 +39,10 @@ def test_user_course_post_failure(client: Client):
     assert r.status_code == 500
     d = client.delete('/api/session', json={'sessionID': sessID})
     assert d.status_code == 200
+
+def test_course_post_not_authorized(client: Client):
+    '''
+    Test user course post without user session/login
+    '''
+    r = client.post("/api/user/course", json=TEST_USER_COURSE)
+    assert r.status_code == 403


src/api/tests/test_user_course.py
```diff
@@ -0,0 +1,40 @@
...
+    r = client.post("/api/user/course", json=TEST_USER_COURSE)
+    assert r.status_code == 200
+    g = client.get("/api/user/course", json=TEST_USER_COURSE)
+    get_data = g.json()[0]
+    assert get_data['course_name'] == TEST_USER_COURSE['name']
+    assert get_data['semester'] == TEST_USER_COURSE['semester']
+    assert get_data['crn'] == TEST_USER_COURSE['cid']
+    d = client.delete("/api/session", json={"sessionID": sess.json()['content']['sessionID']})
+    assert d.status_code == 200
+def test_user_course_post_failure(client: Client):
```
lhain08, line 30: Please leave a single white line between functions

src/api/tests/test_user_course.py
```diff
@@ -0,0 +1,40 @@
...
+def test_user_course_post_success(post_user, client: Client):
+    '''
+    Test user course post by comparing it to user get course
+    '''
+    sess = client.post("/api/session", json=TEST_USER)
+    assert sess.status_code == 200
+    r = client.post("/api/user/course", json=TEST_USER_COURSE)
+    assert r.status_code == 200
+    g = client.get("/api/user/course", json=TEST_USER_COURSE)
+    get_data = g.json()[0]
```
lhain08, line 24: How is this assuring that your course is the first course in the results?

lhain08: On second thought, this is probably fine, other test cases also rely on a fresh database so you can leave it as this.

src/api/tests/test_user_course.py
```diff
@@ -0,0 +1,40 @@
+from .util import Client
```
lhain08, line 1: Please add one more test case for the 'Not Authorized' case.
This should be pretty straight forward, you just need to try to post a new course without logging in (calling session post) and assert you get 403 in response.

lhain08 (changes requested):
Please address the comments and then this should be good to merge!

PR #575, opened 2022-04-20
Tortoise conversion user event

hhdtc:
user event


src/api/db/userevent.py
```diff
@@ -5,7 +5,23 @@ class UserEvent(Model):
...
+    async def addEvent(self, uid, eventID, data, timestamp):
+        sql = """INSERT INTO public.user_event (event_id, user_id, content,created_at) VALUES ('%s', '%s', '%s', '%s');"""
         args = (eventID, uid, data, timestamp)
-        return self.db.execute(sql, args, False)[0]
+        
+        userEvent = await self.db.execute(sql, args, False)
+        return userEvent[0]
+
+    async def updateEvent(self, uid, eventID, data):
+        sql = """INSERT INTO public.user_event (event_id, user_id, content,created_at) VALUES ('%s', '%s', '%s', '%s');"""
```
lhain08, line 16: You are just overwriting this sql string so please delete line 16

src/api/models/__init__.py
```diff
@@ -1,5 +1,7 @@
 from .user_account import UserAccount
 from .user_session import UserSession
+from .user_event import UserEvent
+from .event import Event
```
lhain08, line 4: This is already in the init file, remove line 4

src/api/tests/test_user.py
```diff
@@ -39,7 +39,9 @@ def test_user_post_success(post_user, client: TestClient, event_loop: asyncio.Ab
     '''
     add a valid new user into the session
     '''
+    r = client.post("/api/session", json=TEST_USER)
```
lhain08, line 42: This shouldn't be here, we are creating a new user so signing into any user here does not make sense.

lhain08 (changes requested):
You need to remove user_event.py from the tables directory.

lhain08 (approved):

PR #576, opened 2022-04-20
Tortoise conversion course selection

YuetingLiao:
**Issue**

Convert to tortoise for course-selection
Did not implement and test pytest for course-selection


.gitignore
```diff
@@ -10,3 +10,4 @@ node_modules/
 *.ipynb*
 courses20.xml
 .coverage
+src/api/tests/Dockerfile
```
lhain08, line 13: This should not be getting ignored here. If you did this to stop from pushing some kind of change to the dockerfile, you should prevent that by only adding files you wish to change.

src/api/db/student_course_selection.py
```diff
@@ -1,45 +1,49 @@
-class student_course_selection:
-	def __init__(self, db_conn):
-		self.db_conn = db_conn
+from db.model import *
+from models.student_course_selection import StudentCourseSelection
```
lhain08, line 2:
```suggestion
```

This import is not necessary

src/api/db/student_course_selection.py
```diff
@@ -1,45 +1,49 @@
...
 					DELETE FROM
 						student_course_selection
 					WHERE
-						user_id = %s AND
-						semester = %s AND
-						course_name = %s AND
-						crn = %s
+						user_id = '%s' AND
+						semester = '%s' AND
+						course_name = '%s' AND
```
lhain08, line 38: I could be wrong but I believe its possible for course names to have single quotes within them. We should probably escape this with the $<tag>$...$<tag>$ syntax as a precaution. You can see examples of this in the new db/courses.py file.

src/api/models/student_course_selection.py
```diff
@@ -0,0 +1,16 @@
+from tortoise import fields
+from tortoise.models import Model
+
+class StudentCourseSelection(Model):
+
+    user_id = fields.ForeignKeyField('user_account.user_id')
+    #user_id = Column(INTEGER, ForeignKey('user_account.user_id'))
```
lhain08, line 7: Please remove commented code

src/api/app.py
```diff
@@ -260,16 +260,29 @@ async def log_out(request: Request, session: SessionDeletePydantic):
...
-async def remove_student_course(request: Request, courseDelete:CourseDeletePydantic):
+@app.post('/api/user/course')
+async def add_student_course(request: Request, credentials: UserCoursePydantic):
     if 'user' not in request.session:
         return Response("Not authorized", status_code=403)
-    resp,error = course_select.remove_selection(courseDelete.name, courseDelete.semester, request.session['user']['user_id'], courseDelete.cid)
+    resp, error = await course_select.add_selection(credentials.name, credentials.semester, credentials.cid)
     return Response(status_code=200) if not error else Response(error, status_code=500)
+#
+#
```
lhain08, line 270: You need to finish/uncomment the delete and get methods. It also appears that there was some conflicts here as you have outdated versions of the methods as they are still using flask syntax in the commented versions. So make sure you get the correct version from TortoiseConversion when you do uncomment.

lhain08 (changes requested):
I left a few comments, please let me know when you are done with them and I will take another look.

.gitignore
YuetingLiao: Change has been discard

lhain08: This is still in the files changed

src/api/db/student_course_selection.py
```diff
@@ -1,45 +1,48 @@
...
 					WHERE
-						user_id = %s AND
-						semester = %s AND
-						course_name = %s
+						user_id = '%s' AND
+						semester = '%s' AND
+						course_name = $course$%s$course$
 					"""
-			resp, error = self.db_conn.execute(sql, [uid, sem, name], False)
+			resp, error = self.db.execute(sql, [uid, sem, name], False)
```
lhain08, line 29: Missing await here

lhain08 (changes requested):

lhain08 (approved):

PR #601, opened 2022-10-25
Design of what the new feature should look like

ArnavPawar:
Began design for the new feature
Collaborated with group to produce a design
IMG_1348
IMG_1349


DarknessUnder:
I think you should seprate the branches

Issue #597, opened 2022-10-18
Feature Request — Degree Templates Tab


ethanz2003:
**Job**
Make a section on the page that displays degree templates grouped by school.

**Describe a possible solution**
Use scraped data from degree templates and display on the Yacs site in a similar style to
the Hass Pathways section.

PR #624, opened 2022-11-26
Degree Templates 

ethanz2003:
**Issue**

> closes Feature Request — Degree Templates Tab #597

**Photos**

**Before**
image

**After**
image

**Sorting by Academic Schools**
image

**Sorting Alphabetically**
image

**Additional Info**

Code has a similar implementation to the Pathway section. Has yet to include data from interdisciplinary majors such as BCBP and Interdisciplinary Science.


DarknessUnder (dismissed):

zhengw5 (changes requested):
Everything looks good and works with no bugs, but when you click on Free elective or HASS elective, it brings the user to a page and says course was not found. Clicking on this should either bring us to a page with all possible HASS or Free electives, or it should just not bring us anywhere and act as just some text.

zhangn7 (commented):
Just a suggestion: You could link the explore page when you click on Science/Math options in the curriculum template. Since you link direct courses link when a specific course is on the curriculum, I think you could link the explore page with the MATH/SCIENCE classes they could take.

PR #633, opened 2022-12-09
Select/Remove Course in Explore page

charleschae12:
Select/Remove Course in Explore page function added.

**Issue**

**Select Course in Explore page**
-Added a "Select" button in the 
explore page to add the course 
easily to the “Selected Courses”
-Afterwards, can check through 
“Selected Courses” located in the 
sidebar of the first screen, and 
can easily choose wanted section.

**Remove Course in Explore page**
-After press the "Select" button, 
it changes to "Remove" button.
-Added "Remove" button which 
removes the selected course
-If the course searched are already selected, 
the "Remove" button appears first.

chrome-capture-2022-11-9

### DIFF
--- a/src/web/src/pages/CoursePage.vue
+++ b/src/web/src/pages/CoursePage.vue
@@ -28,8 +28,13 @@
           {{ courseObj.description }}
         </b-col>
       </b-row>
-      <b-button @click="$router.go(-1)">Back</b-button>
+      <b-button @click="$router.go(-1)">Back</b-button> &nbsp;&nbsp;&nbsp;
       <!--      :to="'/explore/' + courseObj.department"-->
+
+      <b-button class="openbtn"
+                :class="{active:isActive}"
+                @click="toggle()"
+                >{{isActive ? 'Remove' : 'Select'}}</b-button>
     </div>
     <CenterSpinner
       v-else-if="isLoadingCourses"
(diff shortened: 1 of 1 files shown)


jonha1 (dismissed):
- well-made feature and easy to use
- couldn't find any bugs

mlammon11 (changes requested):
The course selection button on the explore page works as intended, however, I believe it should provide more feedback to the user.
Some suggestions are:
- change the background color of the button to green/red for selected/removed
- add "successfully added/removed" next to the button

zhengw5 (approved):
-The feature runs with no bugs
-The feature also seems viable

chavab0809 (changes requested):
I found one bug when reviewing the pull request. If you click on the select button for a course and then go to a different page (like the schedule page etc.) and then go back to that selected course it will not say remove, instead it says select again even though you have already selected it.

https://user-images.githubusercontent.com/123212886/226734581-297b4e54-742a-4785-8964-fb57940a775e.MOV

zhangn7 (changes requested):
As chavab said before, once you leave the current page with the course you selected and then go back to that page to remove it, it will not say remove anymore. It requires you to click "select" and then remove instead of only needing to click remove since the course is already selected.

zhengw5 (changes requested):
I agree with chavab and nancy, I checked it myself.

PR #648, opened 2023-01-30
Update docker-compose.production.yml

DarknessUnder:
Testing, round 2


DarknessUnder:
Did not work, wrong direction

PR #667, opened 2023-03-17
Display courses in courselisting in section order

Troy-bailan-King:
courselist


Troy-bailan-King:
And also show professor name for each section

TroYxiaoqinchun (approved):
Overall it is good, courses are ordered ascendingly. Professors' family names are shown.

PR #668, opened 2023-03-17
Qtssuisz

### LINKED ISSUE #653: Feature Request — filtering classes with their corresponding  credit
**Job**
Search class by the credit

adding a filter function to filter classes by how many credit this class is

**Context**

Qtssuisz:
> closes #653

filtering classes with their corresponding credit 

image

### DIFF
--- a/src/web/src/components/CourseList.vue
+++ b/src/web/src/components/CourseList.vue
@@ -134,22 +134,10 @@ export default {
       courseList: null,
       debounceTime: 300,
       Items: [
-                {
-                    name: '1 Credit',
-                    id: 1
-                }, 
-                {
-                    name: '2 Credits',
-                    id: 2
-                }, 
-                {
-                    name: '3 Credits',
-                    id: 3
-                }, 
-                {
-                    name: '4 Credits',
-                    id: 4
-                }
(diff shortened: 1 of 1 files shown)


chavab0809 (approved):
Everything looks good I just ran it on the local hosts and did some tests and they all seemed fine. The only thing I would recommend is formatting the check box next to the credit options to be a little more spaced out. But other than that it seems good!

Qtssuisz:
> chavab0809

I couldn't find a perfect way for the checkboxes layout with space between the checkboxes and their labels but I swapped their position

I hope this will look better

image

DarknessUnder (approved):

ajwadcs (approved):
Runs fine and looks pretty good for an initial design. Maybe could turn the credit filter into a dropdown menu instead of checking boxes, but just a suggestion.

zhangn7 (approved):

mlammon11 (changes requested):
Feature works as intended and is a great addition to improving user experience. However the spacing seems a bit off, if you could fix the spacing or maybe make the buttons into a drop down, then it would be great.

Issue #660, opened 2023-02-17
Feature Request — Flexible height timetable


skyjune0824:
Currently height of the table is fixed, so if the text is too long, they often hidden behind the table.

The table should be expanded as the text kept added.

PR #679, opened 2023-04-04
Update ScheduleEvent.vue

TroYxiaoqinchun:
**Issue**
#660
Some courses are one-hour long, and related time block will be too small to show full information. Now it is scrollable.

Nothing changed here

Test Procedure
> 1. Go to main page
> 2. Select ADMN-1824, or other 1-hours long course
> 3. Scroll the block

**Photos**

https://user-images.githubusercontent.com/97756261/229914431-c352638d-933b-4a82-866d-fba3426cffd3.mp4

**Additional Info**

### DIFF
--- a/src/web/src/components/ScheduleEvent.vue
+++ b/src/web/src/components/ScheduleEvent.vue
@@ -96,6 +96,7 @@ export default {
     }
   }
   .event-text::-webkit-scrollbar {
+    background: var(--sb-track-color);
     width: 8px;
   }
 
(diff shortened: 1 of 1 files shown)


chavab0809 (dismissed):
Just tested it out on multiple one hour courses and all are able to scroll.

zhengw5 (changes requested):
I think the function works really well and is practical however the text color and bold is different from the master branch. I don't think you should change that part of the scheduling.

skyjune0824:
I like the feature! However it could be easily missed by the user, since it is not visible whether the schedule is scrollable or not.

PR #685, opened 2023-04-11
666 schedule disappearing when deselecting course

littlebizaizi:
close issue 666.

change the way it find out conflict courses.
So, now it would just treat a course with no section been selected as a blank course.

### DIFF
--- a/src/web/src/pages/NewCourseScheduler.vue
+++ b/src/web/src/pages/NewCourseScheduler.vue
@@ -504,7 +504,7 @@ export default {
           Object.values(this.selectedCourses)
         );
         if (!result.length) {
-          throw new Error("conflict!");
+          throw new Error("conflict!222");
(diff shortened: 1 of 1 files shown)


jonha1 (commented):
Can you describe an example where this occured

littlebizaizi:
After choosing some courses, if there's a course that only contain 1 section inside, unselect that section (not the course) from the "Selected Course" menu.
At this time the whole schedule would just disappear instead of showing the remaining course been selected.

mlammon11 (changes requested):
This fixed some issues but didn't resolve others. For example, if you add two courses to schedule, then deselect all sections for both courses, then try to add another class - the 3rd class won't be shown on the schedule, and will only show once re-selecting one of the two original courses. If you could look into this that would be great.

PR #694, opened 2023-04-14
When click the course block, lead directly to the course info page

charleschae12:
When click the course block, lead directly to the course info page issue finished

Animation

### DIFF
--- a/src/web/src/components/ScheduleEvent.vue
+++ b/src/web/src/components/ScheduleEvent.vue
@@ -6,7 +6,7 @@
     :style="data.style"
     :id="props.crn + props.day + props.startTime"
   >
-    <a :href="$options.getExploreCourseLink(props.name)">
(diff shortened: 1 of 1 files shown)


zhengw5 (changes requested):
This function is really nice however I found an issue with it. When you do click on a course, it did bring me to the correct page, but if you click on the tab schedule again, the schedule disappears until you change something about your selected courses. This problem persists whenever you leave the schedule, and then click the schedule tab again. So I am not sure if it is problem that this feature caused or if this was just an old version.

jonha1 (commented):
Great feature, but when the user goes back to schedule through the schedule button on the top, schedule disappears unless action was performed upon the schedule. It does work if the user clicks the "back" button.

charleschae12:
The issue is not because of my function, current YACS has the same issue too. But, I will take a look at it if I could fix it.

PR #727, opened 2023-06-13
Path way modal

### LINKED ISSUE #719: Bug — Front End Pathway Page
**Describe the bug**
The popup window for pathways is transparent which causes the contents of the window to the top of the pathway contents causing it too be unreadable and messy.
**To Reproduce**
1. Go to the Pathways navigation bar at the top of the page
2. Click on any of the pathways

**Expected behavior**
The pop-out window should not be transparent and you should be seeing the contents behind it

**Screenshots**
image
**Additional context**

liw20:
changed class type in modal component

closes #719


tenzinpalden (approved):
Fixes the bug as intended

Issue #113, opened 2020-04-21
Ops — Create Pull Request template


jshom:
**Job**
When we are merging pull requests we want to notify all important changes so that the team can work faster to merge new features.

**Describe a possible solution**
Create a pull request template

**Context**
A few issues that affected the front end did not have a clear test plan, in affect, we though it is a great idea to create a template that prompts people to provide such test plan.

PR #741, opened 2023-06-30
Faq dl

doinglin:
**Issue**

Note which issue number(s) this Pull Request closes. All pull requests should close an issue, if one does not exists, make one. This helps us plan and discuss before we decide on implementation

*Example:*
> closes #113

If you added/modified a table, notify people of schema change here

*Example:*
> Added table `account_semester_selection` for a mapping from a students account and semester to a class and section (aka class/section selections)

**Test Modifications**

Note added unit/integration tests: (all backend changes should contain unit/integration tests with pytest)

*Example:*
> Added test `tests/api/db/test_classinfo.py:test_crn_uniqueness` to ensure all classes returned from API call don't have repeat CRNs

**Test Procedure**

*Example:*
> 1. Create a branch
> 2. Commit a change to that branch
> 3. Push branch
> 4. Create Pull Request and fill in PR template

**Photos**

Show before and after, capture screenshot/gif of finished feature/bug

**Additional Info**

More info that would help people review your change

### DIFF
--- a/src/web/src/pages/FAQs.vue
+++ b/src/web/src/pages/FAQs.vue
@@ -2,42 +2,67 @@
     <div class="faq-page">
       <h1>Frequently Asked Questions</h1>
   
-      <div v-for="(question, index) in faqData" :key="index" class="faq-item">
-        <h3 @click="toggleAnswer(index)">{{ question.title }}</h3>
-        <p v-if="question.isOpen">{{ question.answer }}</p>
+      <div v-for="(question, index) in faqData" :key="index" class="faq-item" :class="{ 'active-item': question.isOpen }">
+        <div class="question-wrapper">
+          <h3 @click="toggleAnswer(index)">{{ question.title }}</h3>
+        </div>
+        <div class="divider"></div>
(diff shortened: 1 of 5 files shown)


src/web/src/pages/FAQs.vue
kurodj, line 1: 1.  I think it's better to keep answer_symbol  uniform (symbol for first question is A. where as the rest are A:)

2. For the first answer to "How can I use the Schedule feature?", it should be "at" the top of page rather than "on"

3. For the first answer to "What is the Explore feature?", it should be "at" the top of page rather than "on"

4. For the first answer to "How can I use the Pathways Navigation feature?", it should be "at" the top of page rather than "on"

5. For the second answer to "How can I use the Finals feature?", it should be "finish" rather than "finishing"

kurodj (changes requested):
Few grammatical errors. Everything else works fine.

liw20 (dismissed):
There seems to be a lot of negative space on the sides. But it get the mission and message of yacs across.

PR #764, opened 2023-07-28
Schedule sticky

### LINKED ISSUE #728: Feature Request — Make the open/close button for the class search stay with the class search
**Description**
The open/close button for the class search moves with the schedule instead of with the class search section. If someone wants to close the class search, they would have to scroll back up to close it. It also gets hidden when the window is smaller.
Screenshot 2023-06-16 at 2 35 01 PM
Screenshot 2023-06-16 at 2 35 42 PM

**Solution**
Move the button next to the class search tab.

liw20:
Closes #728
Fixed issue of the close tab button disappearing when scrolling down the page by attaching it to the course search side panel and fixing it to the top left hand corner when the panel is closed. Kept the previously implemented burger-button animation.

### DIFF
--- a/src/web/src/components/Header.vue
+++ b/src/web/src/components/Header.vue
@@ -6,6 +6,7 @@
     toggleable="md"
     type="primary"
     variant="light"
+    sticky="{ boundary: true }"
   >
     <b-navbar-brand
       class="align-middle text-dark"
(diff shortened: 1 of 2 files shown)


mlammon11 (changes requested):
Changes don't improve the website. the button should not overlap the calendar. one idea is to make it a pull tab along the side. not sure if this is even an "issue" that needs to be changed.

PR #779, opened 2023-08-11
New finals page mlammon11

### LINKED ISSUE #778: Feature Request — Improve YACS with this feature
**Job**
Current finals page has some features that need to be added such as:
- deleting course
- search on select
- not preventing search if box is not filled out
- filtering out already selected courses from available list
- etc

image

Possible Solution:
- force user to select course using a modal which eliminates null selection and allows to search after selection

mlammon11:
> closes #778 

OLD
image

NEW
image
image
image
image

*Example:*
> 1. Go to Finals Page
> 2. Add Courses
> 3. Delete Courses
> 4. View available list which excludes already selected courses

### DIFF
--- a/src/web/src/pages/FinalExamScheduler.vue
+++ b/src/web/src/pages/FinalExamScheduler.vue
@@ -15,13 +15,28 @@
                     v-model="selectedCourses[index]"
                     :options="courseOptions"
                 ></b-form-select>
-                <b-button @click="$bvModal.show('delete-modal-id')" variant="danger" class="ml-3">Delete</b-button>
+
+              </b-form-group>
+
+            </div>
+            <b-button @click="addCourse" variant="primary">Add Course</b-button>
+
+            <b-button type="submit" variant="success" class="ml-3">
+              Search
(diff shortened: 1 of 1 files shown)


sophiachen646 (approved):
The feature works as intended, allowing the addition and deletion of finals. The page is done really nicely for easy access and convenience when changing a course. My only suggestion is that the delete button might be confusing for some people since they don't know what it is that is being deleted until they click on it. It might be better to label it 'Delete a course' and possibly put it below the 'Add Course' button so people know that it doesn't automatically delete their added courses.

dorian451 (changes requested):
Looks great so far!
Two things: 
1. Instead of having a singular delete button, maybe have a per-class delete button, like an x in the top corned, above the course title?
2.  The buttons in the dialogs in this page have inconsistent styling compared to the rest of the website. Every other button is filled in, whereas these buttons only fill in when hovered on.

c-jimmy1:
Attached are images of the updated UI addressing the requested changes.

image
image

PR #810, opened 2023-10-02
Course conflict message

jonha1:
**Issue**
Whenever there is a course conflict, the scheduler is empty


jonha1:
Looks great! Love the simplicity.

PR #849, opened 2023-11-28
801 Rewrite Webscraper to use SIS

### LINKED ISSUE #801: Feature Request — Rewrite webscraper to use SIS login
**Job**
The SIS course catalog has a lot more info in a far more standard way when viewing while logged in using an RCSID. If we could just prompt for login info from whoever is running the script we could allow any maintainer to use their own account to generate a csv file with all of the courses without having to go through the registrar.

Part 1. Send a request to sis.rpi.edu, orompt the user for their RCSID and their password.
Part 2. Get the DUO code which shows up, give the user the code
Part 3. Navigate to Course Search, select the correct term on the drop-down.
Part 4. For every subject in the course search subject dropdown, select the subject, parse all courses on the page, and then go back to the course search page.

**Context**
Our current solution is bad.

bnavac:
**Issue**

Fixes #801 

**Test Procedure**

Run new_parse.py


dorian451 (approved):
The fix intended for users that don't use the duo app by default doesn't work but other than that it's fine

jonha1 (approved):
Looks good to me! Great work.

Issue #860, opened 2024-01-26
Feature Request — Make the webscraper interact with our API


martig7:
**Job**
Right now each new info csv needs to be uploaded manually to the website, which is annoying.

**Describe a possible solution**
Make it make an api call every time the script updates the csv. I've been told this is the solution I don't know...

**Context**
Dorian told me to do this

PR #865, opened 2024-02-14
Webscape update

### LINKED ISSUE #864: Bug — Webscraper Pre/Coreq Scraping Broke
At some point early this semester something subtle changed in sis and now all prerequisites and co requisites scraped are incompatible with our csv upload checker. I plan to fix this on the web scraping side by making the prereq scraping on SIS more robust generally. This is a band-aid fix to make sure we can scrape Summer 2024 before it comes out.

martig7:
**Issue**

closes #864 

I updated the web scraper to properly scrape coreqs and prereqs. It should run similarly to before, except parse_runner may or may not need some updating, which will come with the resolution of issue #860

**Test Modifications**

Tests should be the same as before, just run new_parse.py with any given semester and a test csv will be outputted.

**Additional Info**
I also added a specific bugfix for a semester from 2021.


dorian451 (changes requested):
Looks good so far, but it doesn't work on my machine :)
More specifically, it can't sign in to SIS reliably

dorian451 (commented):
Looks good so far (with the changes I did, look at them and make sure they're good)

dorian451 (approved):

dorian451 (approved):

PR #868, opened 2024-02-27
Auto updater

martig7:
**Issue**

> Closes Issue #860 

**Changes**

> Added a new file, post_test.py, which contains a function which makes an upload to our api. This requires the use of an environment variable named "yacs_url", which will just contain the first part of the url that the request is sent to. Additionally updated headless_login to be more consistent and updated parse_runner to work with the changes I made.

**Testing**

> Run parse_runner.py. It should do a parse cycle and then print OK 200 when a file is uploaded.


martig7:
Stale

PR #886, opened 2024-09-24
fix semester deletion

Tevetron:
this change links our semester_info data table and our courses data table so that deleting a semester from our database removes the corresponding class information


becausej (commented):
Don't completely understand, but I've confirmed it doesn't break yacs

Tevetron:
I made an additional change to properly create the foreign key relationship between the semester_info and course tables.

PR #893, opened 2024-11-05
Finals database support

Tevetron:
Add finals database support for new web scraper. It includes uploading a finals schedule, deleting a semester's schedule, and getting finals by semester. Deleting semester and course information at the same time was also merged into this branch.


dorian451 (changes requested):
Can you please get rid of the migrations that do nothing and add testing steps

dorian451 (approved):
seems to work

PR #894, opened 2024-11-15
Makes no_login.py work with command line arguments and automatically send results to the api

becausej:
Added login in no_login.py to take in the year and semester as command line arguments and created send_courses_to_api.py which does what it says it does and it run automatically from no_login.py upon it's completion. There is a no_login_documentation.txt file which describes use now. 

Location: \yacs.n\rpi_data\modules


Tevetron (approved):
looks good to me!!

dorian451 (changes requested):
Please get rid of your .idea folder

